PR #6, opened 2024-06-10
Jax bkg

### LINKED ISSUE #2: The background quantities,  $H(z)$, $\chi(z)$

marcobonici:
In this PR, which closes  we have added the following functionalities:
- Split the Cosmology in a Background and Perturbations backends
- Implemented the within the JAX Backend.
Regarding the latter, it was checked that `CAMB` and `JAXBackground` agree to each other. Furthermore, it was checked that we can progagate gradient calculations in all of these steps. An example of the chains running using `NumPyro` and `JAXBackground` can be found in the `notebooks/sampling` folder.


cloelite/cosmology/jax_cosmology.py
```diff
@@ -0,0 +1,276 @@
...
+        zs : array_like
+            Redshifts at which to calculate the transverse comoving distance.
+
+        Returns:
+        --------
+        np.ndarray
+            The transverse comoving distance as a function of redshift.
+        """
+        return self.comoving_distance(zs)
+    #TODO add the lax conditionals to account for the curvature!
```
gcanasherrera, line 121: Can you open a task about this? CAMB Background backend will have it

PR #42, opened 2024-10-10
Jax Fishers proof of concept

santiagocasas:
Very rough draft on photometric observable fishers using cloelite and jaxcosmo.
Just as a proof of concept, of the things that need to be changed to be able to jit(grad) the whole thing.

Despite merging possible, this should not be merged in this state, until whole implementation and logic is revised.


gcanasherrera:
@santiagocasas I was thinking that now this can be a notebook in `playground`. Opinions?

santiagocasas:
I don't think it's possible @gcanasherrera . I  did this PR long time ago, when there were no protocols and the JAX cosmology class was pretty different. I had done some manual fixes to ensure differentiability like forcing variable types and so on. I think now with the protocols and the new classes, this should have been fixed (are we sure differentiabilty is ensured all the way down @marcobonici ?)
On top of that I had used jaxcosmo as the code to generate the Gaussian Covariance from the lensing C_ells and I did a lot of mixed stuff. Finally, since the Fishers are just the second derivatives of the likelihood, I basically did that directly and compared against first derivatives of the C_ells.
It would be actually quite nice now to do what we said in our discussions and write a [cloelike](https://github.com/cloe-org/cloelike) which is differentiable and uses the Euclid inputs and showcase Fishers from there.
Long story short, I think one needs to completely rewrite this if one wants to use it as simple notebook in playground.
Maybe you can close this PR or archive it and I go back to this once I have some time?

cloelite/cosmology/jax_cosmology.py
```diff
@@ -409,8 +412,8 @@ def linear_matter_power_spectrum(self, ks, zs, **kwargs):
             and scale factor.
 
         """
-        ks = np.atleast_1d(ks)
-        zs = np.atleast_1d(zs)
+        ks = jax_atleast_1d(ks)
```
santiagocasas, line 415: for instance @marcobonici  and @gcanasherrera  I had done these jax_atleast_1d arrays here to ensure the output was jax arrays, but all this is deprecated now thatnks to the Protocols with duck typing

gcanasherrera:
Ok, let's open an issue in cloelike so that we don't forget in the future.

PR #68, opened 2025-01-13
GCspectro first implementation

AndreaPezzotta:


marcobonici:
Amazing, I'll give some comments in the afternoon!

marcobonici:
HI @AndreaPezzotta , as said in person, I need to read it carefully before giving a long comment.
My first suggestion is to split everything in an abstract class (`LegendreMultipoles`) and a some child classes (`Comet`, `PBJ`, `Effort` etc...). General methods, like the AP and the mask convolution, should be put in the parent class, while the computation of the $P(k)$ should be in the child classes.

I'll put some more precise comments on your code!

cloelite/cosmology/camb_cosmology.py
```diff
@@ -3,7 +3,8 @@
 from cloelite.cosmology.cosmology import LinearPerturbations
 from cloelite.cosmology.cosmology import NonLinearPerturbations
 # General imports
-import numpy as np 
+import numpy as np
+from scipy.interpolate import interp1d
```
marcobonici, line 7: This is a general comment, for the moment I don't wanna modify this (I'll take care of it in a second phase!).
In order to ensure we can use `jax`, we need to use specific packages, like `interpax` (repo [here](https://github.com/f0uriest/interpax).) This will ensure we can do gradient computation.
Again, it is just a comment that does not required any action now (let us focus on implementing basic things first!) but I just want to start adding comments about this :)

cloelite/observables/spectro.py
```diff
@@ -0,0 +1,327 @@
...
+class LegendreMultipoles(ABC):
+
+    def __init__(self, NLmodel: Optional[str] = 'EFT',
+                 linear_perturbations: Optional[LinearPerturbations] = None,
+                 background_fiducial: Optional[Background] = None):
+        r"""Class constructor
+        Parameters
+        ----------
+        NLcode: str
+            Non-linear code used to compute the power spectrum
```
marcobonici: Remove these two lines from the docstring.
```suggestion
```

cloelite/observables/spectro.py
```diff
@@ -0,0 +1,327 @@
...
+        for i,ell in enumerate(ells):
+            multipoles[f'ell{ell}'] = \
+                integrate.simps(self._Pk2d_tot(self._k_AP(k, self.mu_grid,
+                                                          parameters['z'],
+                                                          use_AP=use_AP),
+                                               self._mu_AP(self.mu_grid,
+                                                           parameters['z'],
+                                                           use_AP=use_AP),
+                                               params) *
+                                legendre(ell)(self.mu_grid),
```
marcobonici, line 269: In this moment, the AP integration is not done in the most efficient way:
1. The integrand is even, so we might do the integral between 0 and 1 and double the result. This would improve the AP computation efficiency by a factor of 2.
2. We are using simpson, but higher level integration rules might be used (Gauss-Lobatto, Gauss-Legendre or similar). This will reduce the number of points used to do the integral by a factor of 4-5.

Later we will improve it, but for the moment is fine.

marcobonici (changes requested):
I left a couple of minor comments. Some are just comments, and there should be a couple of small things (e.g. lines to remove in docstrings).
After this, we are good to go!

cloelite/observables/spectro.py
```diff
@@ -0,0 +1,327 @@
...
+                         background_fiducial=background_fiducial)
+
+        from comet import comet
+        self.comet_inst = comet(model=NLmodel, use_Mpc=True,
+                                bias_basis='AssBauGre')
+
+    ######### FOR TESTING #########
+    def set_fiducial_cosmology_comet(self, parameters: dict):
+        self.comet_inst.define_fiducial_cosmology(params_fid=parameters)
+
```
marcobonici, line 350: Should this be here?
Since the background fiducial is passed o the constructur of the `LegendreMultipolesComet` object, I suggest to do it when creating the object.

marcobonici:
Another small thing: in this exact moment there is this thing being done in the `spectro.ipynb`

`mps_spectro.comet_inst.define_nbar(nbar=1e-4)`

But the Abstract class has the method `set_number_density` .
My suggestion is to redefine the `set_number_density` method for the `Comet` class, such that it does what it is already doing and also what you need for `comet`.
To be more concrete


```
def set_number_density(self, nbar: float):
    r"""Setter of the number density
    Parameters
    ----------
    nbar: float
        Number density of the considered GCsp sample
    """
    self.nbar = nbar
    self.comet_inst.define_nbar(nbar=nbar)
```

cloelite/observables/spectro.py
AndreaPezzotta: Done

cloelite/observables/spectro.py
AndreaPezzotta: I fixed point 1, as it was straightforward. As for 2, I agree that something like Gauss-Legendre is definitely more optimised. Let's leave it outside of this pull request for the momemt, but I will focus on this, say, after the Garching meeting

cloelite/observables/spectro.py
AndreaPezzotta: I left those functions in the section labelled as "for testing" just because I wanted to check that the multipoles obtained with cloelite were consistent with the native ones produced by comet. Eventually this function and the `power_multipoles_comet` one will be removed. Btw this is not affecting in any way the output of cloelite, since the shot-noise correction is added based on the `self.nbar` attribute, whihch is set using a different function

AndreaPezzotta:
@marcobonici Thanks for the comments! I pushed a new commit in which I fixed few of the points you have raised.

Related to the testing functions, they are just meant to check the consistency of cloelite with comet; if you prefer I can still edit those lines that you mentioned, but those functions will anyway be removed soon. 

With this commit, I also added the extra functions for the convolved power spectrum multipoles, the masking module (for the scale cuts), and the short likelihood module that I used to run the chain I shared few days ago. If this is not required for this pull request, I can remove it.

One last thing, after we discussed on what input the individual non-linear codes should receive from cloelite, I removed the input linear perturbations object from the call to comet. Basically, the only external (to comet) thing that is currently used are the background classes, to compute the AP parameters. For this, I commented the lines in the `update` method of the `camb_cosmology` class. In this way, when we pass new parameters to the `power_multipoles` function, the code is much faster (which is what I needed to run the example chain). In order to run the spectro notebook, you should first pull the comet repo and pip install again from the dedicated branch, since I made some changes there.

cloelite/input/reader.py
marcobonici, line 1: I am hpayy with this, but in Garching we should discuss with Emiliano et al. on how to include these pieces in `euclidlib` (cc'ing also @gcanasherrera ).

PR #70, opened 2025-01-15
gh-69: add pre-commit config

### LINKED ISSUE #69: Code style etc.
Since more PRs are starting to come in, would it be good to harmonise the code style from the beginning?

Standard practise at the moment is to set up pre-commit so that everything is checked locally on developer's machines, and then use the pre-commit CI service to check also PRs with the exact same rules.

You can see examples in https://github.com/glass-dev/glass but also a lot of other places. Something to keep in mind for the future?

ntessore:
Add a pre-commit config. For the time being, only ruff and ruff-format are run, plus a few standard formatting checks.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,33 @@
+---
+ci:
+  autofix_prs: false
+  autoupdate_commit_msg: "pre-commit.ci: update pre-commit hooks"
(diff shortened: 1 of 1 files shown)


ntessore:
If the maintainers are happy with the pre-commit setup, you can use https://pre-commit.ci to get the same checks on PRs.

Issue #53, opened 2024-12-05
Deal with installation of external emulators + data


gcanasherrera:
Investigate the best approach to make sure that the emulators are installed if cloelite is installed, jointly with the data, to prevent issues in clusters that do not have connection to internet.

itutusaus:
If I may add, I'd suggest to also ensure the installation of all packages that are used in the notebooks we provide as examples, to make sure they run out of the box

gcanasherrera:
There is a newest issue opened by @chiaramoretti this one is deprecated now.

PR #72, opened 2025-02-11
gh-49: play with protocols for background, perturbations, tracer classes

### LINKED ISSUE #49: Play with protocols for Background & Perturbations classes
Re-shuffling of the Background & Perturbations classes to make them as agnostic as possible of the ingested code

gcanasherrera:
### Description:
This PR replaces abstract base classes (ABCs) with Protocol to improve flexibility and future compatibility with future external codes (interfaces). 

### Changes:

- Replaced ABCs abstract classes with Protocols
- Removed difference between LinearPerturbations and NonLinearperturbations, with protocols both classes are compatible
- Create a protocol for Tracer

### Why?

- Reduces unnecessary inheritance constraints.
- Allows external implementations to adhere to the expected interface without modifying their class hierarchy.
- Improves compatibility with other codes

### Next Steps:

- Validate integration with different backend implementations (#45 , #53 ).
- Add unit tests

Let me know your thoughts! 🚀


ntessore:
Hi @gcanasherrera, I will look at the protocols in more detail soon, but first a comment on the constants: why take them from astropy, instead of writing these O(10) floats down? By using an external source of constants, a user risks changing every single CLOE result when they update astropy.

gcanasherrera:
> Hi @gcanasherrera, I will look at the protocols in more detail soon, but first a comment on the constants: why take them from astropy, instead of writing these O(10) floats down? By using an external source of constants, a user risks changing every single CLOE result when they update astropy.

I see your point, you’re absolutely right. I’ll make that change in a follow-up.

ntessore:
Looking at the proposed design, I have two big-picture comments.

### Background and Perturbations split

I think the split into `Background` and `Perturbations` as a separate interface makes life more difficult than it needs to be. Consider the following alternative:

```py
class Cosmology(Protocol):
    # shared methods and attributes for all cosmologies
    ...

class MatterPowerCosmology(Cosmology, Protocol):  # ignore the name, just an example
    # additional methods that you need for matter power spectra
    ...
```

In this design, an instance of `MatterPowerCosmology` is also a `Cosmology`, so you do not need to pass two instances around, just one. Why is this good? Because it means you do not have to figure out whether any specific method (e.g. `sigma_8`) lives in `cosmo` or `perturb` — it's always in the object you are given. Another advantage is that `cosmo` and `perturb` can never be inconsistent.

In practice, this also makes it much easier to implement e.g. your emulators, since you only need to "patch" a few methods on top of an otherwise fully functional instance. If you want to keep building the "perturbation" around an existing "background" instance (but see below), you only have to wrap that object. For example, consider this:

```py
class CosmologyWrapper:
    def __init__(self, cosmology: Cosmology) -> None:
        self.cosmology = cosmology

    @property
    def H0(self) -> float | Array:
        # forward to contained cosmology instance
        return self.cosmology.H0

    ...
```

You can then implement an emulator for the matter power spectrum basically exactly as you are doing now:

```py
class MyMatterPowerEmulator(CosmologyWrapper):
    def __init__(self, cosmology: Cosmology, *args, **kwargs) -> None:
        super().__init__(cosmology)  # set up the wrapper

    # implement only the matter power methods here
    def matter_power_spectrum(self, ...) -> ...:
        # get sigma_8 from cosmology
        sigma_8 = self.cosmology.sigma_8
        ...
```

But `cosmo = MyMatterPowerEmulator(...)` is subsequently a fully-featured `MatterPowerCosmology` instance, which means it's also a `Cosmology` instance, and you only need to pass that one object around (without worrying about consistency).

### Perturbations keeping background instances

I believe the headache in designing your interfaces all come from one fundamental issue: The current approach of implementing "extensions" (either a single one as above, or all of "perturbations" as currently) on top of a "background" has the drawback that you cannot easily implement multiple of these "extensions" at the same time. An instance has to have every possible functionality of "perturbations", and there is no way to create an implementation that only provides, e.g., an emulator in one place without also providing every other method.

A more powerful and flexible way to go would be using what's called "mixins". Instead of having a class hold a "background" and then implement some additions on top of that, you could implement the above matter power emulator like this:

```py
class MyMatterPowerEmulator:  # no superclass
    # !!! no init

    # implement only the matter power methods here
    def matter_power_spectrum(self, ...) -> ...:
        # get sigma_8 from cosmology
        sigma_8 = self.sigma_8  # !!! this calls sigma_8 on self
        ...
```

You can then create a _new_ cosmology class with the matter power emulator on top of any given cosmology/background class like this:

```py
class MyCosmology(MyMatterPowerEmulator, CambCosmology):
    """Combines the MatterPowerEmulator with a CAMB cosmology."""
```

This new class that combines an extension with the base cosmology is created by the user, not in the library (for prior art, see e.g. Python's [`socketserver`](https://docs.python.org/3/library/socketserver.html#socketserver.ForkingMixIn)). Here, `MyCosmology` is implementing the full `MatterPowerCosmology` protocol from above. And you could easily add a `BaryonicFeedbackEmulator` on top, or any other orthogonal extension.

gcanasherrera:
Hello @ntessore,  

Thanks a ton for your insightful feedback! I’ve been carefully exploring different approaches, especially your last suggestion about mixins—an interesting concept I wasn’t familiar with, so I really appreciate your explanation!  

I do see the advantages of consolidating Background and Perturbations into a single `Cosmology` class, particularly in terms of flexibility, especially if `Cosmology` is structured as a mixin of modular building blocks. The idea of stacking `Boost/InitialConditions/DarkEnergyParametrisation` on top of Background is quite appealing, as it would allow us to construct a fully functional instance using pre-existing components.  

However, at this stage, mixins come with a notable downside: initializing them isn’t always straightforward. I’ve experienced this firsthand, and while a shared parameter object/block could help, it risks resembling the old `cosmo_dic` in CLOE—something that most developers found frustrating. Additionally, using mixins effectively requires a solid understanding of Python’s MRO, which isn’t necessarily intuitive for the average cosmologist who will be working with the code. That said, I think we can overcome these issues by either providing predefined mixins for users to import or ensuring strong documentation (or something fancier!). This is something I’d love to explore further, and I’ll push a preliminary branch with a mixin-based implementation in the near future so we can discuss it in this task #76 

I have thought about the suggestion you also shared privately about keeping the `matter_power_spectrum` a simple method that could be later attached to a pre-defined `Cosmology` class. I see the point that coming with a proper `Perturbations` class is not straight-forward and we might lose flexibility in the near future, but, although most likely you are right, I'm taking the risk for now.

That means, then, I’m sticking with the Background and Perturbations split for a few reasons:  

1. **It works.** While somewhat rudimentary, the current setup integrates well with the external codes we rely on.  
2. **It’s easy to explain.** The separation is conceptually clear, making it accessible to users. It took us 15 minutes to explain to the team what we meant, and the spectro team have integrated the spectro codes very easily.
3. **We need a v1.** Having an end-to-end implementation that allows us to plot Cls and gather feedback will be crucial before redesigning the interface. Reworking the architecture now would introduce significant refactoring overhead, particularly given the ongoing work on the spectro side.  

That said, the shift from abstract classes to protocols remains. Would you be open to reviewing the current implementation? In particular, I’d love your thoughts on whether I’ve completely messed up the numpy/jax compatibility.  

Thanks again for your commitment on this. Really appreciate the detailed breakdown and the stimulating discussion, it made my weekend. Let's move the discussion to #76 as I said before :)

ntessore:
Hi @gcanasherrera. I had a look through the proposed interface, and I think it can work. I'm slightly cautious about the fact that, even though you are covering all use cases now, we might run situations we cannot easily accommodate in the future. But I think we already discussed ways this can be mitigated.

Since this is quite the big change, I think the most efficient way forward is to merge it as-is, and then open a couple of issues about specific points unrelated to the interface design?

ntessore (approved):

gcanasherrera:
@all-contributors please add @ntessore for tool, mentoring, code, review

PR #78, opened 2025-02-20
gh-77: fixing cl integration

### LINKED ISSUE #77: Fixing Cl integration
After some restructuring in #72 , the $C_\ell$ integration looks to be broken.
Investigate and fix this issue.
Relevant also for #54 .

marcobonici:
Hello @gcanasherrera ,
I actually fixed the issues.
Here is a list of the bugs fixed.
1. **Units**. There was a mismatch with the units previously used (hubble against non hubble units).
2. **Interpolation**. The interpolator was giving a silent error (nans) when giving the Limber interpolated Pk. Now it extrapolates without giving nans.
3. **Growth factor.** The growth factor was expecting a different signature. This is something we have to homogeneize: in this moment, the Perturbations protocol assumes that the growth factor takes as input z and k, while the CAMB cosmology does not take inputs.
The target for this PR is your branch.


gcanasherrera:
I've added milestones, tags, and changed the title of the PR to be linked to the task :)

cloelite/observables/photo.py
```diff
@@ -242,9 +250,8 @@ def get_window_positions(self, z) -> np.ndarray:
         window_positions: numpy.ndarray
            Window function for angular photometric galaxy clustering
         """
-
         window_positions = self.dndz * \
-            self.perturbations.background.hubble_parameter(z)/c_0
+            self.perturbations.background.hubble_parameter(z, units = "km/s/Mpc") / c_0
```
gcanasherrera: I would go with the approach of assuming that H(z) is always in "km/s/Mpc", to avoid problems with further interfaces (and remove the units from perturbations.background.hubble_parameter). Agree?

cloelite/summary_statistics/two_point.py
```diff
@@ -20,12 +20,12 @@
 def Cl_integration(WT1, WT2, Pkl, H, chi2):
     # To be updated
     # still have to include weights, basically we are doing unnormalized trapz
-    return np.einsum('iz,jz,lz,z,z->lij', WT1[:, 1:], WT2[:, 1:], Pkl[1:, 1:], 1/H[1:], 1/chi2[1:])
+    return np.einsum('iz,jz,lz,z,z->lij', WT1, WT2, Pkl, 1/H, 1/chi2)
```
gcanasherrera, line 23: We will need to be specifics in the docs that n(z) cannot start from zero. Maybe opening a task about this?

cloelite/summary_statistics/two_point.py
```diff
@@ -20,12 +20,12 @@
...
 
 @jax.jit
 def Pkl_interp(k_l, z_l, ks, zs, Pk):
     return 10**interpax.interp2d(jax.numpy.log10(k_l), z_l, jax.numpy.log10(ks),  zs,
-                                 jax.numpy.log10(Pk), method="cubic")
+                                 jax.numpy.log10(Pk), method="akima", extrap=True)
```
gcanasherrera, line 28: Will this change when #79 is also dealt with? If not, maybe worth remembering arguing why in the paper

cloelite/cosmology/__init__.py
```diff
@@ -1,4 +1,4 @@
-__all__ = ['cosmology', 'jax_cosmology']
+__all__ = ['cosmology']#temporary fix, 'jax_cosmology']
```
gcanasherrera, line 1: This will be tackled in #80

gcanasherrera (commented):
Only minor comments, otherwise, ready to be merged in the protocols branch

cloelite/observables/photo.py
marcobonici: Yes. I'll change that.

cloelite/summary_statistics/two_point.py
marcobonici: I am opening that

marcobonici: Btw: we do agree that this is also a sensible "physical" requirement, right?

cloelite/cosmology/__init__.py
marcobonici: ✅

cloelite/summary_statistics/two_point.py
gcanasherrera: We will need to double check when we get the first SGS n(z) data product, but seems reasonable

cloelite/summary_statistics/two_point.py
marcobonici: No.
This is a separate issue. The Limber, since it has to go to very low z, takes very high values of k. This means that it will always be necessary some degree of extrapolation. But this is not worrysome, since at the end that contribution is killed by the n(z).

cloelite/observables/photo.py
marcobonici: Done with tha last commit.
Ready to be resolved.

cloelite/summary_statistics/two_point.py
gcanasherrera: Worth to put it already in the paper, I'll copy paste this comment

cloelite/summary_statistics/two_point.py
marcobonici: Opened issue #81 .

gcanasherrera:
Ok, we're ready!

marcobonici:
Hi @gcanasherrera , I have changed the default behaviour of H(z), such that the output is in km/s/Mpc if not stated otherwise.
I also opened the tasks you asked for.
Now should be ready to be merged!

Issue #79, opened 2025-02-20
Better lensing efficiency


marcobonici:
Lensing efficiency is slow.
Implement a triangular matrix a' la CosmoCentral.jl, since it will make everything faster.

marcobonici:
Now that the shear has been fixed in #78 , we can move with the second phase: efficiency.

gcanasherrera:
This is now implemented, closing the task.

Issue #74, opened 2025-02-12
clean auxiliary


gcanasherrera:
auxiliary directory is becoming messy, adapt and clean for v1 release

PR #82, opened 2025-02-22
gh-79: better lensing efficiency

marcobonici:
In this PR I have added a fast method for the evaluation of the lensing efficiency, based on a triangular simpson matrix.
The old functions to compute it were not removed, but I simply added a suffix `_check` at the end of the old functions name.
Once we are gonna have unit tests, these can be used to test the accuracy of the new function.

Can you please check and review it, @gcanasherrera ?

### DIFF
--- a/cloelite/auxiliary/math_utils.py
+++ b/cloelite/auxiliary/math_utils.py
@@ -89,7 +89,13 @@ def stacked_simpson(n):
     return jnp.vstack(rows)
 
 #this 100 is clearly hardcoded. Think of a better mechanism!
-precomputed_simpson_matrix = stacked_simpson(100)
+_cached_stacked_simpson = {}
+
+def cached_stacked_simpson(n: int):
+    key = str(n)
+    if key not in _cached_stacked_simpson:
+        _cached_stacked_simpson[key] = stacked_simpson(n)
+    return _cached_stacked_simpson[key]
 
 def legendre(n, x):
     if n == 0:
--- a/cloelite/auxiliary/math_utils.py
+++ b/cloelite/auxiliary/math_utils.py
@@ -2,7 +2,7 @@ import jax
 import jax.numpy as jnp
 
 # This will go to a math-utils module, for the moment stays here
-max_size = 300 
+max_size = 300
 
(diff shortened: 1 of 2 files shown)


gcanasherrera:
@marcobonici Can I start reviewing?

cloelite/auxiliary/math_utils.py
```diff
@@ -2,7 +2,7 @@
 import jax.numpy as jnp
 
 # This will go to a math-utils module, for the moment stays here
-max_size = 300 
+max_size = 300
```
gcanasherrera, line 5: Put it already in math utils, we will re-arrange math utils in #74

cloelite/observables/photo.py
```diff
@@ -170,14 +171,29 @@ def get_lensing_efficiency_bin(self, z, bin_idx):
...
         n_bins = self.dndz.shape[0]
         efficiency = self.get_lensing_efficiency_bin(z, 0)
         for i in np.arange(1,n_bins):
             efficiency = np.vstack([efficiency, self.get_lensing_efficiency_bin(z, i)])
 
         return efficiency
 
+    def get_lensing_efficiency(self, z):
+        dndz = self.dndz
+        dz = z[1]-z[0]#assuming equispaced!
```
gcanasherrera, line 184: We need to start collecting this assumptions somewhere. Open a task about generalising this (so we don't lose track), please

cloelite/observables/photo.py
```diff
@@ -200,6 +216,22 @@ def get_window(self, z):
         """
         return self.get_lensing_window(z) + self.get_window_IA(z)
 
+    def get_window_new(self, z):
```
gcanasherrera: I'm not sure I agree with the names used. Let's give some descriptive names instead of _new.

marcobonici:
> @marcobonici Can I start reviewing?

Sure.

gcanasherrera (commented):
All for you to go!

cloelite/observables/photo.py
marcobonici: I do not agree with the name chosen as well. I will modify it.

Issue #13, opened 2024-06-10
Introduce neutrinos in the background and perturbation classes


gcanasherrera:
At the moment, neutrinos are not include for milestone v0.3. 

In the future, we would like to account for neutrinos.

Implementation: TBD

PedroCarrilho:
@gcanasherrera, I think we can follow up on neutrinos here. 

Note @chiaramoretti already added the neutrino mass as a parameter in her PR to ensure it was set to 0, otherwise CAMB was setting it to the default 0.06. I had put just the default value of 0.06 in the emulator for things to match CAMB and will now add it as a parameter, as Chiara did.

We will need to sort out the modelling for neutrinos everywhere. Things that come to mind:

- Creation of P_cb for linear and nonlinear. This means that `matter_power_spectrum` may require some options depending on what we want.
- Correcting the growth rate (and growth factor) for scale dependence
- Probably changing some things in the PBJ and COMET implementations (but maybe that already works)

AndreaPezzotta:
In addition to the bullets specified by @PedroCarrilho, let me also add this comment. If we decide to work with the assumption of only considering the large-scale value of the growth rate (and considering the impact of neutrinos only at linear level) we should revisit how the growth rate is defined with CAMB.

Right now, we get it as the ratio of fsigma8 over sigma8. This works perfectly fine when not including neutrinos. However, on the other case, the integration over 8Mpc/h scales is affected by the damping of massive neutrinos, and the ratio no longer matches the large-scale ratio between the cross density-velocity power spectrum and the density-density power spectrum (which by definition should be f)

Here's an example. The orange and green lines show the ratios Pdv/Pdd and Pvv/Pdd, respectively, for a cosmology with Mnu=0.5eV. The large-scale limit of these ratios should correspond to f and f^2, respectively. For each color, there are two curves, one obtained by considering the density field of cdm and baryons only, and one where the density field also includes neutrino fluctuations. Irrespectively from which case one choose (traditionally galaxy clustering models build up on top of cdm+b fluctuations), the large-scale limit is the same.

[...]

PR #89, opened 2025-03-07
gh-88: mixing_matrix_3x2_HMemu_fix

### LINKED ISSUE #58: Include the HMCode 2020 emulator

### LINKED ISSUE #88: Multiplication with the mixing matrix for 3x2
Implementation of the mixing matrix multiplication to compute pseudo C_ell.

PedroCarrilho:
I added a first version of the mixing matrix multiplication. Beyond this I fixed a few things that were stopping the HMemu from running. More changes are likely to be required there, but will leave that to future PR. I also added a notebook to compute and plot the pseudo_Cell and test the emulator and its speed.


gcanasherrera:
Hello @PedroCarrilho , I'm going to start with the review, sorry for not seeing it, but it was not linked to any branch.

Just to make github not to go crazy, what are the tasks that this PR would cloe? Only #88 ? Potentially #58 too?

PedroCarrilho:
I suppose it would also close #58, although that could also have been closed with your previous merge. One the other hand, baryons are not yet included, so I guess we can close 58 when I put that in.

PedroCarrilho:
New update: I now made baryons work (we may want to discuss how to pass TAGN, but I just built an example for the emulator). I think with this, we can close #58 too.

cloelib/auxiliary/extrapolator.py
```diff
@@ -8,7 +8,7 @@ def extend_spectra(wavenumber_in, redshift_in, boost_in,
                  extrap_func=None,
                  extrap_kmax=500.0,
                  extrap_kmin=1e-4,
-                 extrap_zmax=4.0,
```
gcanasherrera, line 11: I'm curious about this change in the name...

cloelib/auxiliary/extrapolator.py
```diff
@@ -52,8 +52,8 @@ def extend_spectra(wavenumber_in, redshift_in, boost_in,
         max wavenumber for extrapolation in 1/Mpc
     extrap_kmin: float
         min wavenumber for extrapolation in 1/Mpc
-    extrap_zmax: float
-        max redshift to extrapolate
+    extrap_z: float or np.ndarray
```
gcanasherrera, line 55: I feel the zmax was better

cloelib/auxiliary/extrapolator.py
```diff
@@ -69,7 +69,11 @@ def extend_spectra(wavenumber_in, redshift_in, boost_in,
     """
 
     wavenumber_base = np.geomspace(extrap_kmin,extrap_kmax,500)
```
gcanasherrera: maybe we can promote the 500 and 100 numbers to variables, even hard-coded, but that allows to change in the future

cloelib/cosmology/HMcode2020Emu_cosmology.py
```diff
@@ -38,7 +39,7 @@ def __init__(self, background : Background, redshifts: np.ndarray):
             'As': self.background.As,
             'ns': self.background.ns,
             'hubble': self.background.H0 / 100,
-            'neutrino_mass': 0.0,
+            'neutrino_mass': 0.06,
```
gcanasherrera: I feel we should be consistent with CAMB too.

cloelib/cosmology/HMcode2020Emu_cosmology.py
```diff
@@ -54,9 +55,27 @@ def __init__(self, background : Background, redshifts: np.ndarray):
...
+                           option_cosmo="const", ns=self.background.ns)
+
+        self.k = k_out
+        self.z  = z_out
+        self.Pk = Pk_out
+
+        pk_interp = interpolate.RectBivariateSpline(
+            self.z , self.k, Pk_out, kx=1, ky=1)
+
+        self.Pk_interp = pk_interp
```
gcanasherrera, line 77: I'm still not super convinced about this, which is the reason why I was delaying my review, but I cannot come up with a better approach right now.

cloelib/summary_statistics/angular_two_point.py
```diff
@@ -1,11 +1,13 @@
...
 from cloelib.observables.tracer import Tracer
+from cloelib.observables.photo import PositionsTracer
+from cloelib.observables.photo import ShearTracer
 from cloelib.auxiliary.units import SPEED_OF_LIGHT
 
 # General imports
 import interpax
 import jax.numpy as np
-from scipy.interpolate import RectBivariateSpline
+from scipy.interpolate import RectBivariateSpline, interp1d
```
gcanasherrera: @marcobonici any ideas to keep this part differentiable? interp1d is not

gcanasherrera (changes requested):
First general comments. Overall two main concerns:

- We agreed that `matter_power_spectrum` should be dependent on z,k, as well as the growth factor and growth rate (at least z) CC: @chiaramoretti 
- I'm worried about the number of hardcoded variables. Maybe we can be explicit in the docs of the methods.
- @marcobonici can you have a look at differentiability for the computation of the pseudo_Cls

cloelib/auxiliary/extrapolator.py
PedroCarrilho: This is because we can now pass a full z array or just the z_max. Actually in most cases we will pass the full array, so I thought it best to call it just z instead of zmax.

cloelib/cosmology/HMcode2020Emu_cosmology.py
PedroCarrilho: This change was because the default value we are using in CAMB is 0.06, so this change is making it consistent with CAMB. I could also make this explicit in CAMB or start passing the value of the neutrino mass, as we should do. I believe @chiaramoretti  was doing something like this in her branch, but what do we prefer to do now?

cloelib/summary_statistics/angular_two_point.py
PedroCarrilho: Replaced this with interpax version

cloelib/auxiliary/extrapolator.py
PedroCarrilho: Done

PedroCarrilho:
I think I mostly responded or corrected the things you asked. Let me know if you have any other thoughts

cloelib/cosmology/HMcode2020Emu_cosmology.py
gcanasherrera: To be addressed in #107

cloelib/cosmology/HMcode2020Emu_cosmology.py
gcanasherrera: Let's leave neutrinos aside for the moment

gcanasherrera (approved):
PR approved! We need to address #107 and @PedroCarrilho can you open a generic issue to follow up on neutrinos ? Thanks

After that task is open, I'll merge, simply ping me!

PedroCarrilho:
@gcanasherrera , since I merged Chiara's PR, I will have to resolve some conflicts. I will ping you once that is done.

I also commented on #107 and in #13 for neutrinos. I can either solve some of the things in #107 here or do that in another PR, just let me know what you prefer.

PedroCarrilho:
I resolved the conflicts so everything should now be ready, but let me know what you prefer regarding the other issues.

PR #98, opened 2025-03-21
gh-48: Pbj implementation protocols

### LINKED ISSUE #48: Preliminary implementation of PBJ
TBI after discussion #44 concludes

### LINKED ISSUE #91: Linear growth rate normalisation
Hi @gcanasherrera I spotted a possible bug in `camb_cosmology.py`: when computing the growth rate, we should be normalising it to redshift z=0, but as is the code uses just the first redshift specified at class initialisation. See e.g. [here](https://github.com/cloe-org/cloelib/blob/1642fe0f351641a5d1cc47f073e8c7ecf8267195/cloelib/cosmology/camb_cosmology.py#L228) but the same is true for the same method in `CAMBNonLinearPerturbations`, while the implementation in HMcode seem to be correct (uses the power spectrum interpolator set as class attribute by the `matter_power_spectrum` method).

Possible solution is to do the same as in HMcode_cosmology: in `matter_power_spectrum` we set as class attribute `self.Pk_linear` the camb interpolator, while the method still returns an array (but now it's the interpolator evaluated at z,k)? we then call the interpolator when computing the growth factor

chiaramoretti:
I have implemented PBJ following the protocol structure and the structure used by COMET.
I have also fixed the bug described in #91 for `camb_cosmology.py`, changing the behavior of the `growth_factor` and `matter_power_spectrum` methods of `CAMBLinearPerturbations`, that now take as parameters redshifts and scales

I have not changed these methods in `CAMBNonLinearPerturbations`


cloelib/cosmology/camb_cosmology.py
```diff
@@ -178,25 +180,45 @@ def __init__(self, background: Background, redshifts: np.ndarray) -> None:
...
 
-        self.background.interface_args['CAMBparams'].set_matter_power(redshifts=redshifts, kmax=self.kmax)
+        self.background.interface_args['CAMBparams'].set_matter_power(
+            redshifts=redshifts, kmax=self.kmax)
         self.results = camb.get_results(self.background.interface_args['CAMBparams'])
 
-    def matter_power_spectrum(self) -> np.ndarray:
-        """
-        Calculates the linear matter power spectrum.
+        self.k_values, _, self.Pk_linear = self.results.get_linear_matter_power_spectrum(
```
PedroCarrilho: It would be great if we could have the same variable names everywhere (between camb and emulator and inside every Perturbations protocol, etc). 

I had been using `self.k`, `self.Pk` as well as `self.z`, but we can use your `self.k_values` (and perhaps `self.z_values`?). I would, however use just `self.Pk` instead of `self.Pk_linear`, so that also the nonlinear ones have a `self.Pk` with the same name. ID'ing them would be easy since they are elements of different classes, so there would be no confusion. 
Do you think that is OK? Fine if not, just let me know what you prefer.

cloelib/cosmology/camb_cosmology.py
```diff
@@ -178,25 +180,45 @@ def __init__(self, background: Background, redshifts: np.ndarray) -> None:
...
-            hubble_units=False, k_hunit=False
-        )
-        self.k = k_values
-        self.z = z_values
-        self.Pk_linear = pk_values
+        pk_values = camb.get_matter_power_interpolator(
+            self.background.interface_args['CAMBparams'],
+            nonlinear=False, extrap_kmax=self.kmax,
+            hubble_units=hubble_units, k_hunit=k_hunit,
+            var1='delta_tot', var2='delta_tot').P(zs, ks)
```
PedroCarrilho, line 221: I think here it is preferable to use 
```
pk_values = self.results.get_matter_power_interpolator(
            nonlinear=False, extrap_kmax=self.kmax,
            hubble_units=hubble_units, k_hunit=k_hunit,
            var1='delta_tot', var2='delta_tot').P(zs, ks)
```
since I think the way you have it, it will call camb again from the beginning, instead of using the already computed result (but I could be wrong). This may not make a difference in terms of result for the linear case, but when doing this for nonlinear, I noticed that it does not remember which nonlinear model was defined before and so just uses the default one. Using the function form the `results` instance did not have this problem, so that is probably best. 

In any case, let me know if you had a reason from doing it your way, or change it to this way if you agree.

PedroCarrilho: Another thing here: Would it be useful to define the interpolator in the `__init__` and then just call it here? In 3x2, we actually need to call `matter_power_spectrum` multiple times and the way it is written now, we would be creating different interpolators every time, which might waste (a little) time unnecessarily. If you agree, we could move the interpolator and create something in `__init__` like

```
self.Pk_interp = self.results.get_matter_power_interpolator(
              nonlinear=False, extrap_kmax=self.kmax,
              hubble_units=hubble_units, k_hunit=k_hunit,
              var1='delta_tot', var2='delta_tot').P
```

and then in this function just return `self.Pk_interp(zs,ks)`.

notebooks/observables/spectro_protocols.ipynb
```diff
@@ -122,45 +154,49 @@
...
    "metadata": {},
    "outputs": [],
    "source": [
     "k = np.logspace(-3, np.log10(0.3), 101)\n",
-    "mps = mps_spectro.power_multipoles(k=k, ells=[0,2,4])\n",
-    "mps_noAP = mps_spectro.power_multipoles(k=k, ells=[0,2,4], use_AP=False)"
+    "mps = mps_spectro_comet.power_multipoles(k=k, ells=[0,2,4])\n",
+    "mps_pbj = mps_spectro_pbj.power_multipoles(k=k, ells=[0,2,4])\n",
+    "\n",
+    "mps_noAP = mps_spectro_comet.power_multipoles(k=k, ells=[0,2,4], use_AP=False)"
```
PedroCarrilho: Shouldn't this be in a different cell, for example, right before the AP comparison is plotted? Also, is it also possible to do an AP comparison with PBJ? This is not necessary, just asking.

PedroCarrilho (changes requested):
Everything looks good with code, just had some comments on the CAMB part to make it all more uniform.

The notebook looks good. I left a comment on it, but this is probably not for now, since you and Andrea will probably move all of this to the other repo. Running it did not go very well, so I have a few more comments about that:
- Comet is giving me the error `AttributeError: 'PTEmu' object has no attribute 'Pk2d'`. Perhaps I did something wrong while installing it or have the wrong version? I think I got the latest from pip.
- I ran the parts I could with PBJ, but the plot on the impact of RSD parameters looks different when I replace the comet stuff with the pbj, so not sure what is going on.
image

- Every other plot with pbj is working fine and is reproduced without any issue.

Let me know if some of this is not clear.

PS: To run the notebook, I installed PBJ, but had issues with the required tensor flow version (2.14), since both cosmopower wants <2.14 and (at least on Mac) 2.14 requires ml-dtypes=0.2.0 while jax (at least the version I have) requires 0.4.0. There may be some way to resolve these package conflicts, and this is not for now, but just wanted to point it out for the future. Also, you had mentioned there was a light way to install PBJ but I could not figure it out now.

cloelib/cosmology/camb_cosmology.py
chiaramoretti: I agree! I switched to `self.k`, `self.z`, `self.Pk` in `CAMBLinearPerturbations`, so it matches what you are using

gcanasherrera: Sorry for intervening, but didn't we agree that we wanted to make matter_power_spectrum z and k dependent?

cloelib/cosmology/camb_cosmology.py
chiaramoretti: The `get_matter_power_interpolator` was my initial choice too, but it seems there is something wrong with interpolation: if I initialise a CAMBLinearPerturbation object at some specific redshift (e.g. z=1), when calling the `matter_power_spectrum` method at redshift z=0 I get an error 
```CAMBError: P(z,k) requested at z=0, but only computed for z=[1.]. Cannot extrapolate!```
The issue seems to not be there if I create the CAMBLinearPerturbation instance with more than one redshift, but this clashes with the current structure we set up for the SpectroPower objects, they are supposed to be -- at least for now -- created for a single redshift. How about I open an issue for the future, so that when we change the behaviour of the SpectroPower objects we can use the implementation you suggest?

notebooks/observables/spectro_protocols.ipynb
chiaramoretti: You're right, it's confusing, I moved it to a separate cell right before the AP comparison is done. The AP comparison for PBJ is doable, the reason I did not add it is that the AP is handled by the LegendreMultipoles, so if the no-AP P(k,u) is the same (as shown by the pbj-comet comparison plot) then also the AP is the same. I can add it if you think it's useful

chiaramoretti:
Thanks @PedroCarrilho I implemented one minor change and updated the notebook.
I also reply to the two other points you made:
>   Comet is giving me the error AttributeError: 'PTEmu' object has no attribute 'Pk2d'. Perhaps I did something wrong while installing it or have the wrong version? I think I got the latest from pip.

To access the implementation of `Pk2d` in comet you need to install a specific branch `test_lib`. I'm opening an issue to keep track of this, as I imagine we will need to add some instructions on how to install the external codes (e.g. if specific branches are required)

>    I ran the parts I could with PBJ, but the plot on the impact of RSD parameters looks different when I replace the comet stuff with the pbj, so not sure what is going on.

This should now be fixed in the notebook: PBJ has a separate parameter dictionary where I converted the counterterms in order to match the comet prediction, but the part you are referring to was using the COMET dictionary, hence the differences

cloelib/cosmology/camb_cosmology.py
chiaramoretti: Hi @gcanasherrera the `matter_power_spectrum` method is indeed z and k dependent, I added this in the `__init__` because the previous implementation of `matter_power_spectrum` was setting k, Pk as attributes of the class and I wasn't sure these are needed somewhere else thus removing them could have broken the code

cloelib/cosmology/camb_cosmology.py
PedroCarrilho: Ok, that is fine. In my merge, I will have to use what I suggested for the nonlinear part though, but I think that is fine. Let's continue chatting on #107 .

PedroCarrilho (approved):
All looks good. Will merge and continue discussing on #107. If changes come from there, I will probably implement them on my PR (if we agree to that)

PR #100, opened 2025-03-21
gh-99: Implementation of CLASS Cosmology, closes #99

### LINKED ISSUE #91: Linear growth rate normalisation

### LINKED ISSUE #99: Implementation of Boltzmann solver CLASS
### Description
Create CLASSBackground, CLASSLinearPerturbations and CLASSNonLinearPerturbations Protocols, following the same approach as for CAMB. Make sure the input and output arrays are the same as for CAMB and that it receives the same cosmological parameters.

### Goals/Acceptance Criteria
- [x] **Goal 1**: A notebook showcasing CAMB and CLASS cosmology.

### Steps to Reproduce (if applicable)
Run the cosmology_validation.py notebook.

### Expected Outcome
Same figures for CAMB and CLASS.

### Additional Notes

- I had to fix some broken things in the `cosmology_validation.ipynb` notebook. I understand that notebooks are being moved to the playground, so some of my fixes might have been already resolved.
- The f(z) (growth rate) of CAMB seems to be differently normalized compared to what is usually done in other pipelines. I have used the D(z=0)=1 normalization. This has been reported in #91  and probably has been taken care of by @chiaramoretti .
- I also had to fix the JAXBackground and Perturbations to get the examples running. This probably solves #80. 
- I will leave other small naming inconsistencies, improvements and suggestions for optimization for a later issue.

### Assignees
- @santiagocasas

### LINKED ISSUE #109: Details to fix in Cosmology module
### Description
There are some small inconsistencies to fix in order to get the Cosmology module release-ready. I noticed this in the CAMBCosmology, while implementing them in CLASS.

### Goals/Acceptance Criteria
- [ ] **Goal 1**: Omega_b method, change text in its docstring: baryonic density.
- [ ] **Goal 2**: Homogenize arguments of matter_power_spectrum
- [ ] **Goal 3**: Homogenize arguments of growth_factor and growth_rate
- [ ] **Goal 4**: Agreement in f(z) between CAMB and CLASS.

### Current Behavior (if applicable)

1. Omega_b method, text in its docstring says matter density.
2. CAMBLinearPerturbations power spectrum accepts zs, ks as arguments. The non-linear matter power spectrum does not. This is related to task #107 ?
3. Growth factor accepts zs, ks. While growth_rate does not receive arguments. That is a bit inconsistent and could be cumbersome for external users.
4. Growth_factor returns an array which is essentially k- and z-dependent. The growth rate as its derivative should also be then k-dependent, in principle. (This is done like this in the MP/CF validation). 
5. Growth rate from CAMB does not seem to agree with CLASS (scale_independent_growth_factor) and neither with CosmicFish (validated against camb/class, etc....). I thought this was related to the normalization, but it is not the case. Am I missing something? See plots below and `cosmology_validation.ipynb` notebook.

## CAMB linear perturbations:

Image

## CLASS linear perturbations:

Image

## CosmicFish linear perturbations (from camb or class, sub per-mille agreement)

Image

### Additional Notes

Finally, the agreement is now worse after we have set neutrinos to zero. There are some very well-known issues in the agreement for this sector between CAMB and CLASS. See what we did in https://arxiv.org/abs/2303.09451 and https://arxiv.org/abs/2405.06047.  In summary, we had to define in CosmicFish classy two variables, m_nu_camb and Omega_m_camb (read papers for explanation) in order to get perfect agreement. This is probably not necessary at this stage or even in the next months, but I wanted to point this out before it is forgotten.

### Assignees
Let's discuss which of these details is important and I can assign myself some of the quick fixes. Also, depends on #107.

santiagocasas:
This implementation of CLASS follows the same patterns as for CAMB, with the correct naming and python API of classy.
It closes issue #99 and optionally others mentioned therein.

The `notebooks/cosmology_validation.py` notebook shows the comparison between CAMB and CLASS and shows that the interface has the same input/output between CAMB and CLASS, thanks to the Protocols.

The plots below show the comparison of the quantities.

image

image

image

## These comparisons fail because of Bug #91. But the CLASS implementation (normalisation) should be the most standard used in LSS.

image

image


cloelib/cosmology/class_cosmology.py
```diff
@@ -0,0 +1,316 @@
...
+            The growth factor as a function of redshift and wavenumber.
+        """
+        if hasattr(self, 'Pk_linear') and self.Pk_linear is not None:
+            self.D_z_k = np.sqrt(self.Pk_linear / self.Pk_linear[0, :])
+        else:
+            self.matter_power_spectrum()
+            self.D_z_k = np.sqrt(self.Pk_linear / self.Pk_linear[0, :])
+        return self.D_z_k
+    
+    def f_deriv(self, k_fix=True, fixed_k=1e-3) -> np.ndarray:
```
gcanasherrera: I would personally make this method private... agree?

cloelib/cosmology/class_cosmology.py
```diff
@@ -0,0 +1,316 @@
...
+        self.results.compute()
+
+    @property
+    def _interface_args(self) -> dict:
+        """
+        Save internal structure format of interface codes
+        """
+        return self.interface_args
+    
+    def matter_power_spectrum(self) -> np.ndarray:
```
gcanasherrera: We are making this method z and k dependent after internal discussion. Could you please check PBJ pull request?

cloelib/cosmology/class_cosmology.py
```diff
@@ -0,0 +1,316 @@
...
+        self.interface_args = copy.deepcopy(self.linear_perturbations._interface_args)
+        self.interface_args['CLASSparams']['nonlinear_min_k_max'] = 50
+        self.interface_args['CLASSparams']['hmcode_tol_sigma'] = 1e-8
+        self.interface_args['CLASSparams']['non linear'] = nonlinear_model
+        self.interface_args['CLASSparams']['z_max_pk'] = np.max(self.z)
+        self.results = Class()
+        self.results.set(self.interface_args['CLASSparams'])
+        self.results.compute()
+
+    def matter_power_spectrum(self) -> np.ndarray:
```
gcanasherrera: Same as before.

cloelib/cosmology/class_cosmology.py
```diff
@@ -0,0 +1,316 @@
...
+        if k_fix:
+            # Return the redshift part (the first row), since all k's are equal by construction
+            return self.f_zk[0, :]
+        else:
+            # Return the full 2D array (varying over z and k)
+            return self.f_zk
+
+class CLASSNonLinearPerturbations:
+    def __init__(self, background : Background, 
+                 linear_perturbations: CLASSLinearPerturbations,
```
gcanasherrera: Do not make this class dependent on CLASSLinearPerturbations, could it be self-standing?

gcanasherrera (commented):
Minimal things, thanks for adapting to the protocols.

cloelib/cosmology/class_cosmology.py
santiagocasas: sure, I'll do that

cloelib/cosmology/class_cosmology.py
santiagocasas: yeah, I can check that PR. However, I based my implementation on the `main` branch and the CAMB implementation. So, next time we have to plan a bit better the order of the merges and which branches implement what. We would have to do everything much more atomically to avoid back and forths and merge conflicts.

cloelib/cosmology/class_cosmology.py
santiagocasas: yeah, I can make it dependent on the Perturbations protocol, is that fine?  A bit like the original JAXNonLinearPerturbations was implemented.
Otherwise, if you don't agree with my solution at the moment (to retrieve D and f from linear), I can implement growth and growth rate with the nonlinear power spectrum, but I think leaving the user with these unphysical options is not the best solution.

santiagocasas:
@gcanasherrera  you can review now. This solves #109  and #99. I merged branch `109` into this one, so that branch can be removed.

cloelib/cosmology/class_cosmology.py
gcanasherrera: I agree that unphysical solutions are not the best approach, but I would rather give something unphysical at this stage than messing up with the protocols.

gcanasherrera (approved):
Very neat work @santiagocasas , ready to merge.

PR #101, opened 2025-03-23
gh-74: clean auxiliary

### LINKED ISSUE #74: clean auxiliary

marcobonici:
In this PR, I have removed a couple of import that were not actually used and cleaned the math_utils module.
Specifically, previously a few numbers were hardcoded in the module (regarding the number of points used to compute the simpson integral); now these numbers were removed and a cache mechanism has been put in place.

### DIFF
--- a/cloelib/auxiliary/math_utils.py
+++ b/cloelib/auxiliary/math_utils.py
@@ -1,49 +1,6 @@
 import jax
 import jax.numpy as jnp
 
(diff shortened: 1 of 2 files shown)


gcanasherrera:
Nothing against it, good job @marcobonici and thanks!

PR #108, opened 2025-03-23
Added check on z range for tracer

### LINKED ISSUE #81: Check on n(z)
An n(z) with a lower z exactly equal to zero creates problems, as there is a division by chi(0)=0.
Add a check that enforces this thing to be working.

marcobonici:
Closes #81 .

### DIFF
--- a/cloelib/observables/photo.py
+++ b/cloelib/observables/photo.py
@@ -42,7 +42,7 @@ class ShearTracer:
             A 1-dimensional array representing the redshift values corresponding to the `dndz` array.
         """
         if 0. in z:
-            raise ValueError("One of the z array element is equal to zero, breaking Limber integration.")
+            raise ValueError("One of the z array elements is equal to zero, breaking Limber integration.")
         self.perturbations = perturbations
         self.background = self.perturbations.background
         self.dndz = dndz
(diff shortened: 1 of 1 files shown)


marcobonici:
This is a pretty straighforward review, very compat.
The main thing is the error message. Right now it is `One of the z array element is equal to zero, breaking Limber integration.`. Any suggestion on how to improve it?

cloelib/observables/photo.py
```diff
@@ -41,6 +41,8 @@ def __init__(self, perturbations: Perturbations, dndz: np.ndarray, z: np.ndarray
         z : np.ndarray
             A 1-dimensional array representing the redshift values corresponding to the `dndz` array.
         """
+        if 0. in z:
+            raise ValueError("One of the z array element is equal to zero, breaking Limber integration.")
```
santiagocasas: @marcobonici there is a small gramatical error. It should be: "One of the z array elements is equal to zero, breaking Limber integration."

santiagocasas (commented):
All looks good to me, thanks!
This should catch as well the weird edge case that there is a 0 (int) in the array. I think that would also be true because of the np.ndarray.

I posted the grammatical mistake in your sentence.
Another way of writing the error message (a bit more verbose and cleaner) could be:
"One of the elements in the z array is equal to zero, which breaks Limber integration."

Thanks!

cloelib/observables/photo.py
marcobonici: I have pushed a new commit, fixing the error...thx @santiagocasas !

PR #115, opened 2025-03-25
gh-113: Multiplication_with_mixing_matrix_for_GCspectro

### LINKED ISSUE #113: Multiplication with mixing matrix for GCspectro
## 📝 Issue Summary
The function implementing the convolution with the mixing matrix for GCspectro is deprecated after the merging of the new protocl structure

- [ ] Fixing signature of function and potentially anything broken in the body
- [ ] Adding example to spectro notebook

### ✅ Expected Outcome
The `convolved_power_multipoles` function in the `LegendreMultipoles` class will no longer be broken, and an example of its use will be shown in the spectro notebook

### 🐞 Current Behavior
The current function accepts as arguments the parameter dictionary containing all the parameters (cosmological and nuisance) necessary to build a theory model. With the merging of the new protocol, this is no longer necessary, since the parameters are passed from the init of the class or from its lower-level classes (like Background for cosmology and SpectroPower for RSD parameters) 

### 👥 Assignees
- @AndreaPezzotta 

Addressed in #115

AndreaPezzotta:
Closes #113


cloelib/summary_statistics/legendre_multipoles.py
```diff
@@ -228,52 +228,45 @@ def power_multipoles(self, k: np.ndarray,
...
+            Flag to switch between with and without AP corrections
         Returns
         -------
         multipoles_out: dict
             Convolved power spectrum Legendre multipoles
         """
+        ells_tot = [0, 2, 4]
+        ells = self._ensure_array(ells) if ells is not None else ells_tot
 
         self.mixing_matrix_dict = mixing_matrix
```
chiaramoretti: Is it really needed to re-assign the mixing matrix as attribute each time we call this function? Since it's more data related than theory related, I would leave it as parameter of the function without setting it as attribute, unless it's needed somewhere else

chiaramoretti (commented):
All good here, I only left a minor comment. I do have one additional question: the `if all(np.array_equal(kin_arrays[0], kin) for kin in kin_arrays):` is needed to take care of a case where the multipoles are binned differently, but I'm not sure that could ever happen...am I mistaken?

cloelib/summary_statistics/legendre_multipoles.py
AndreaPezzotta: No it is indeed not necessary. This is just a leftover of my original implementation in which I was defining the mixing matrix as class attribute but only once in the init of the class. Until we don't have clear how to interface the GCspectro data handling with the reader, I would remove this line, and reconsider adding it in a future commit but not inside the function itself to avoid doing the assignment at each step

chiaramoretti: great, indeed we'll need to revise this later on when we have the datamodel for the mixing matrix, but for now it works

AndreaPezzotta:
Thanks for the review @chiaramoretti ! I pushed a new version with an edit to address one of your comments. As for the different k binning of the different multipoles, this was originally implemented like this because of the structure of the GCspectro benchmark file for the ISTL runs. While it should be possible to have a different binning, realistically this is probably not gonna happen. If you agree, I would leave the check there for the time being, and update it another time when we'll have a better understanding of the LE3 products, but really, your call

chiaramoretti:
All good from my side, we can come back to this once we know more about the format as you say! Merging :heavy_check_mark:

PR #122, opened 2025-03-30
gh-10: transverse distance and lax conditional

### LINKED ISSUE #10: Transverse distance and lax conditional
Right now we are computing the transverse distance assuming the effect from curvature is only within the comoving distance calculation. In order to do it properly, we need to incldue the additional factor (sin/sinh,identity) according to the value of the curvature. In order to make it JAX compatible, we have to code it with the lax conditional. Include it.

CAMBBackrgound already offers this functionality.

marcobonici:
The goal of this PR is to fix #10, in order to have `JAXCosmology` working with the protocol interface.
Furthermore, in order to support curvature, we have also added `lax_conditionals`, in order to make it jax compatible.

### DIFF
--- a/cloelib/cosmology/jax_cosmology.py
+++ b/cloelib/cosmology/jax_cosmology.py
@@ -53,6 +53,25 @@ class JAXBackground:
         self.gamma_MG = gamma_MG
         self.mnu = mnu
 
+        # Initialize JaxBgk parameters
+        self.interface_args = {'JAXparams': {}}  # Use a dictionary for CLASS parameters
+        self.interface_args['JAXparams']['H0'] = self.H0
+        self.interface_args['JAXparams']['Omega_b'] = self.Omega_b0
+        self.interface_args['JAXparams']['Omega_cdm'] = self.Omega_cdm0
+        self.interface_args['JAXparams']['Omega_k'] = self.Omega_k0
+        self.interface_args['JAXparams']['n_s'] = self.ns
+        self.interface_args['JAXparams']['m_ncdm'] = self.mnu
+        self.interface_args['JAXparams']['A_s'] = self.As
+        self.interface_args['JAXparams']['w0_fld'] = self.w0 # or w0
+        self.interface_args['JAXparams']['wa_fld'] = self.wa # or wa
+
+    @property
+    def _interface_args(self) -> dict:
+        """
+        Save internal structure format of interface codes
+        """
+        return self.interface_args
+
     def hubble_parameter(self, zs, units: str = "km/s/Mpc") -> np.ndarray:
         """
         Returns the Hubble parameter as a function of redshift.
@@ -147,6 +166,30 @@ class JAXBackground:
             The angular diameter distance as a function of redshift.
         """
         return self.transverse_comoving_distance(zs)/(1+zs)
+
+    def Omega_b(self, zs: np.ndarray) -> np.ndarray:
+        """
+        Returns the baryon density as a function of redshift.
+
+        Args:
+            zs (np.ndarray): Array of redshifts.
+
+        Returns:
+            np.ndarray: Matter density values.
+        """
+        return np.array([self.Omega_b0 * (1+z)**3 /(self.hubble_parameter(z)/self.H0)**2  for z in zs])
+
+    def Omega_m(self, zs: np.ndarray) -> np.ndarray:
+        """
+        Returns the matter density as a function of redshift.
+
+        Args:
+            zs (np.ndarray): Array of redshifts.
+
+        Returns:
+            np.ndarray: Matter density values.
+        """
+        return np.array([(self.Omega_b0+self.Omega_cdm0) * (1+z)**3 /(self.hubble_parameter(z)/self.H0)**2  for z in zs])
 
 class JAXLinearPerturbations(Perturbations):
     def __init__(self, background : Background):
--- a/cloelib/cosmology/jax_cosmology.py
+++ b/cloelib/cosmology/jax_cosmology.py
@@ -65,9 +65,21 @@ class JAXBackground:
             Hubble parameter values at specified redshift(s).
 
         """
-        return self.H0 * np.sqrt((self.Omega_b0+self.Omega_cdm0)*np.power(1+zs, 3) +
+        c_0 = SPEED_OF_LIGHT / 1000
+        x = self.H0 * np.sqrt((self.Omega_b0+self.Omega_cdm0)*np.power(1+zs, 3) +
                                  (self.Omega_k0)*np.power(1+zs, 2) +
                                  (1-self.Omega_b0-self.Omega_cdm0-self.Omega_k0) * np.power(1+zs, 3*(1+self.w0+self.wa))*np.exp(-3*self.wa*zs/(1+zs)))
+
+        def default_case(x):
+            return x
+
+        def one_Mpc_case(x):
+            return x/c_0
+
+        conditions = np.array([units=="km/s/Mpc", units=="1/Mpc"])
+        index = np.argwhere(conditions, size=1).squeeze()
+
+        return lx.switch(index, [default_case, one_Mpc_case],x)
 
     def comoving_distance(self, zs: np.ndarray) -> np.ndarray:
         """
@@ -83,7 +95,7 @@ class JAXBackground:
         np.ndarray
             The comoving distance as a function of redshift.
         """
-        c_0 = c_0 = SPEED_OF_LIGHT / 1000  # Convert to km/s
+        c_0 = SPEED_OF_LIGHT / 1000  # Convert to km/s
         fun = lambda x: 1/self.hubble_parameter(x)
 
         def myquad(x, fun):
--- a/cloelib/cosmology/jax_cosmology.py
+++ b/cloelib/cosmology/jax_cosmology.py
@@ -114,9 +114,7 @@ class JAXBackground:
         conditions = np.array([self.Omega_k0>0., self.Omega_k0<0., self.Omega_k0==0.])
         index = np.argwhere(conditions, size=1).squeeze()
 
-        return lx.switch(index,
-                     [positive_case, negative_case, default_case],
-                     p)
+        return lx.switch(index, [positive_case, negative_case, default_case],p)
 
     def angular_diameter_distance(self, zs) -> np.ndarray:
         """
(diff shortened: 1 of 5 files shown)


gcanasherrera:
Quick comment before I take off: this implementation is not compatible with the Perturbations protocol. The matter power spectrum method needs to be called `matter_power_spectrum`. Also, the other methods from Perturbations need to be implemented. Thanks for the effort @marcobonici

marcobonici:
Hi @gcanasherrera , in the original issue we were discussing only the background (although you are right, I was not referring specifically to Background here).
The goal here is to implement only the protocol for the background part.

Additionally: why did you change the name of the PR to gh? What does gh stands for?

cloelib/cosmology/jax_cosmology.py
```diff
@@ -19,41 +20,51 @@
 
 """
 
-class JAXBackground(Background):
-    def __init__(self, H0: float, Omb: float, Omc: float, Omk: float, As: float, ns: float,
-                 w: float, wa: float, sigma8: float, gamma_MG: float):
-        r"""
+class JAXBackground:
+    def __init__(self, H0: float, Omega_b0: float, Omega_cdm0: float, Omega_k0: float,
+                 As: float, ns: float, w0: float, wa: float):
```
santiagocasas: Though we will probably not implement neutrinos or gamma_MG in Jax, I would suggest to add there the positional arguments for those parameters as well (as we have in CLASS/CAMB) since it is then easier to test and run these modules for instance by passing the same tuple.  We probably will need to move away from positional arguments in the future, but for now I would unify.

cloelib/cosmology/jax_cosmology.py
```diff
@@ -19,41 +20,51 @@
...
 
+        Args:
+            H0 (float): Hubble parameter in [km/s/Mpc].
+            Omega_b0 (float): Baryonic matter density parameter.
+            Omega_cdm0 (float): Cold dark matter density parameter.
+            Omega_k0(float): Curvature density parameter.
+            As (float): Scalar amplitude of primordial fluctuations.
+            ns (float): Scalar spectral index.
+            w0 (float): Equation of state parameter for dark energy.
+            wa (float): Time evolution of the dark energy equation of state.
```
santiagocasas, line 40: add here the other positional arguments as well

cloelib/cosmology/jax_cosmology.py
```diff
@@ -19,41 +20,51 @@
...
+        self.h = self.H0 / 100
+        self.Omega_b0 = Omega_b0
+        self.Omega_cdm0 = Omega_cdm0
+        self.Omega_k0 = Omega_k0
+        self.As = As
+        self.ns = ns
+        self.w0 = w0
+        self.wa = wa
 
     def hubble_parameter(self, zs) -> np.ndarray:
```
santiagocasas: In the CAMB/CLASS backgrounds we have defined here a units keyword argument with default "km/s/Mpc", please add as well ;)

cloelib/cosmology/jax_cosmology.py
```diff
@@ -67,51 +78,44 @@ def comoving_distance(self, zs) -> np.ndarray:
...
-        -------
-        Matter density parameter: numpy.ndarray
-            Matter density as a function of redshift
+        Args:
+            zs (np.ndarray): Array of redshifts.
 
+        Returns:
+            np.ndarray: Transverse comoving distance values.
         """
+        c_0 = SPEED_OF_LIGHT / 1000  # Convert to km/s
```
santiagocasas: you are not using this c0 variable inside this function

santiagocasas (changes requested):
Thanks for the PR, Marco!
In order for the JAXBackground to respect the Protocol structure, you should also add a Omega_n and an Omega_m method, on top of another attribute called _interface_args (see CLASS implementation). Only then, the `isinstance(obj, protocol)` method will yield True.
You should probably open another PR as well and implement quickly the JAXLinearPerturbations and NonLinear, in order to complete the jax_cosmology module.
Cheers!

santiagocasas:
@marcobonici  for my PR #123  I branched out of this, because I had to fix the cosmology protocol itself. So, you might want to pull my #123  branch and start from there and do these small fixes, so that your Background protocol is up-to-date with the rest.

cloelib/cosmology/jax_cosmology.py
marcobonici: Ok, done!
Now, a suggestion: how to include the neutrino masses?
The proper way would be to account for the transition between relativistic/non relativistic species. However, the goal of this module is illustrative (and as a check that jax stuff is working!), so I don't expect this to require perfection, cosmology wise.

Any suggestion?

cloelib/cosmology/jax_cosmology.py
marcobonici: Done.

cloelib/cosmology/jax_cosmology.py
marcobonici: Done.

cloelib/cosmology/jax_cosmology.py
marcobonici: Removed.

marcobonici:
Hi @santiagocasas , I have addressed your comments.
The last, missing bit, IMHO, is to decide how to deal with neutrinos. I don't think it is particularly relevant, but still I wanna get to a common conclusion.
My two cents is to include them purely as non-relativistic matter (also given what is there for Eisenstein-HU), but if we disagree we can discuss it. After we agree, I'll address the second round of comments and check that I do get the same results for background quantities as with CAMB-CLASS.

cloelib/cosmology/jax_cosmology.py
santiagocasas: yes, go ahead witn non-realtivistic!  it will be enough for late-time LSS and as a proof of concept!

marcobonici: Added with my latest committ.

cloelib/cosmology/jax_cosmology.py
```diff
@@ -65,10 +65,22 @@ def hubble_parameter(self, zs, units: str = "km/s/Mpc") -> np.ndarray:
...
+        def default_case(x):
+            return x
+
+        def one_Mpc_case(x):
+            return x/c_0
+
+        conditions = np.array([units=="km/s/Mpc", units=="1/Mpc"])
+        index = np.argwhere(conditions, size=1).squeeze()
+
+        return lx.switch(index, [default_case, one_Mpc_case],x)
```
santiagocasas, line 102: great!

cloelib/cosmology/jax_cosmology.py
```diff
@@ -115,7 +115,6 @@ def transverse_comoving_distance(self, zs: np.ndarray) -> np.ndarray:
         Returns:
             np.ndarray: Transverse comoving distance values.
         """
-        c_0 = SPEED_OF_LIGHT / 1000  # Convert to km/s
         delta_z = self.comoving_distance(zs)
```
santiagocasas, line 138: good!

cloelib/cosmology/jax_cosmology.py
```diff
@@ -148,6 +167,30 @@ def angular_diameter_distance(self, zs) -> np.ndarray:
...
+        Returns the matter density as a function of redshift.
+
+        Args:
+            zs (np.ndarray): Array of redshifts.
+
+        Returns:
+            np.ndarray: Matter density values.
+        """
+        return np.array([(self.Omega_b0+self.Omega_cdm0) * (1+z)**3 /(self.hubble_parameter(z)/self.H0)**2  for z in zs])
+
```
santiagocasas, line 195: Perfect!

cloelib/cosmology/jax_cosmology.py
```diff
@@ -85,9 +85,10 @@ def hubble_parameter(self, zs, units: str = "km/s/Mpc") -> np.ndarray:
 
         """
         c_0 = SPEED_OF_LIGHT / 1000
-        x = self.H0 * np.sqrt((self.Omega_b0+self.Omega_cdm0)*np.power(1+zs, 3) +
+        Omega_m0 = self.Omega_b0+self.Omega_cdm0+self.mnu/(93.14*(self.H0/100)**2)
+        x = self.H0 * np.sqrt(Omega_m0*np.power(1+zs, 3) +
                                  (self.Omega_k0)*np.power(1+zs, 2) +
-                                 (1-self.Omega_b0-self.Omega_cdm0-self.Omega_k0) * np.power(1+zs, 3*(1+self.w0+self.wa))*np.exp(-3*self.wa*zs/(1+zs)))
+                                 (1-Omega_m0-self.Omega_k0) * np.power(1+zs, 3*(1+self.w0+self.wa))*np.exp(-3*self.wa*zs/(1+zs)))
```
santiagocasas, line 91: good!  
Just thinking about the perturbations already, do you have a fitting formula for them in the E&Hu case?
Or the plan is to treat them only here as an extra non-rel species?

santiagocasas (commented):
Looks all good to me!  
I will proceed to merge this into my branch (test it as well) and then you can review that yourself.

cloelib/cosmology/jax_cosmology.py
marcobonici: Yes, this is the plan. Later I might even include the Eisenstein-Hu for massive neutrinos or maybe this [symbolic regression](https://arxiv.org/abs/2410.14623).

santiagocasas: Also, in the future we might want to have a separate class for the neutrino sector. This is related to #13 .
In the CAMB/CLASS validation we did, we had to introduce explicitly this 93.14eV factor in CLASS, as it is not hard-coded but rather calculated from the energy-momentum distribution. This depends as well on N_eff and the degeneracy factors and all that, so we need to be very careful about it and it's better to separate it.
Just adding this comment here for the future, but for now it's all good.

santiagocasas: yeaah, I was waiting for you to say it! That symbolic regression is already [included in Cosmicfishpie ](https://cosmicfishpie.readthedocs.io/en/latest/cosmicfishpie.cosmology.html#cosmicfishpie.cosmology.cosmology.boltzmann_code.symbolic_results), as I was working with these UK guys on it last year. It is suuuper easy to port this into any jax code, if I wasn't that slow I'd had it already.

santiagocasas:
Everything is working perfectly!  As agreed, and for convenience I have merged this branch into #123 and the commits of @marcobonici  will be included there. 
JAX is now a proper protocol!  Thanks @marcobonici !!

image

Issue #96, opened 2025-03-17
Add constant bias nuisance parameters


gcanasherrera:
### Description
Include constant bias nuisance parameters to position tracer

### Goals/Acceptance Criteria
- [ ] **Goal 1**: TBA

### Steps to Reproduce (if applicable)
TBA

### Current Behavior (if applicable)
TBA

### Additional Notes
TBA

### Assignees
- TBA

PedroCarrilho:
Hi @gcanasherrera. Just pushed a first version of this to branch 96-photo_bias. It all seems to be working, but may still need some discussion, specially regarding jax as I went with the simplest option of using standard if statements (cc @marcobonici )

I branched this from the mix mat branch so this can go after that, but I won't open a pull request yet.

gcanasherrera:
So, what do we do with this?

Issue #76, opened 2025-02-18
Modify Background + Perturbations + Tracers for mixins


gcanasherrera:
- Explore the use of mixins to define a parent cosmology protocol
- Explore use of mixins for defining nuisance systematic effects classes for Tracers (IA, bias models... )

PR #127, opened 2025-04-03
gh-96: Adding linear photometric galaxy bias

PedroCarrilho:
I have added the 3 options available in CLOE for the linear photometric bias:

- `per_bin`: 1 bias value applied as a constant within each redshift bin
- `per_bin_int`: 1 bias values per bin, applied at the maximum of the n(z) and interpolated in redshift
- `poly`: 3rd order polynomial 

In all cases, the biases are produced as an array and are multiplied at the level of the window. For `per_bin` this is along the bin axis, while in the other cases it is along the z axis. This is different from how it was done in CLOE, since there were other intermediate power spectra that were created (P_gg and P_gm) to account for this. That will be necessary for NL bias, so we could also do it here already (unless we create additional windows for all necessary NL terms). 

Beyond this, there are a number of if statements that I used to select which model to use. I think from what Marco said and showed in jax_cosmology.py, I can replace them now, but let me know if you would like me to do that now or if we leave that for later.


cloelib/cosmology/camb_cosmology.py
```diff
@@ -285,7 +285,7 @@ def __init__(self, background: Background, redshifts: np.ndarray,
         """
 
         self.background = background
-        self.kmax = 100
+        self.kmax = 500
```
marcobonici, line 288: Is there a specific reason to get to such high k? Stability of the interpolator/extrapolation?

cloelib/observables/photo.py
```diff
@@ -237,7 +237,8 @@ def get_window(self, z):
         return self.get_lensing_window(z) + self.get_window_IA(z)
 
 class PositionsTracer:
-    def __init__(self, perturbations: Perturbations, dndz: np.ndarray, z: np.ndarray):
+    def __init__(self, perturbations: Perturbations, dndz: np.ndarray, z: np.ndarray,
+                 galaxy_bias_model: str, nuisance_params: dict):
```
marcobonici, line 242: For the moment this is fine, but the way I envision it is by creating a class `galaxy_bias_model` with different possible implementations.
Then, rather than including the bias in the window, the `galaxy_bias_model` instance is used with a `perturbation` instance to compute the $P_{gg}$. From this point on, the calculations proceed as usual.

Again: this is just to start brainstorming, we don't have to do this now :)

cloelib/observables/photo.py
```diff
@@ -268,8 +267,24 @@ def __init__(self, perturbations: Perturbations, dndz: np.ndarray, z: np.ndarray
...
+        self.n_z_bins = dndz.shape[0]
+        if self.flags['galaxy_bias_model'] in ['per_bin']:
+            self.bias_array = np.asarray([nuisance_params['b1_photo_bin%d'%bin] for bin in range(self.n_z_bins)])
+        elif self.flags['galaxy_bias_model'] in ['per_bin_int']:
+            self.bias_array = np.asarray([nuisance_params['b1_photo_bin%d'%bin] for bin in range(self.n_z_bins)])
+            self.bias_array = interpax.interp1d(self.z,self.z[np.argmax(dndz,axis=1)],self.bias_array,extrap=True)
+        elif self.flags['galaxy_bias_model'] in ['poly']:
+            poly_order = 3
+            self.bias_array = np.asarray([nuisance_params['b1_photo_poly%d'%bin] for bin in range(poly_order+1)])
+            self.bias_array = self.bias_array[0] + self.bias_array[1] * z + self.bias_array[2] * z ** 2 + + self.bias_array[3] * z ** 3
```
marcobonici: This to me looks error prone, but this is due to the way nuisances are dealt with now, so I don't see an easy fix that will not change the higher level architecture.
I think this will be dealt in an easy manner once we change the way we deal with nuisance.
As before: this is more for start brainstorming.

marcobonici (commented):
Hi @PedroCarrilho ,
besides what I wrote in the comments, that is more meant to be a discussion, I am happy with what you wrote.
I dunno whether @gcanasherrera wanna give a second look, I am ok with this PR but I'd like to start thinking about how to improve the way we deal with nuisances (and maybe open an issue, @gcanasherrera :eyes: )

cloelib/observables/photo.py
gcanasherrera: I agree with Marco (I was already envisioning that we should go with classes for the IA too). Question: can we still do jaxy `if` similar to `jax_cosmo...`? Otherwise, we are breaking the hax on this side, after the effort to have a jax example working.

gcanasherrera:
Good morning gentlemen,

Thanks for your work! I agree with Marco overall, only one comment from my side about the `ifs`.

About start chatting about the nuisances, @marcobonici , what if we start brainstorming here #76 ?

marcobonici:
Let me take care of that, in the same manner I dealt with it for the transverse comoving distance.

cloelib/cosmology/camb_cosmology.py
PedroCarrilho: Yes, this actually makes a very substantial difference. See below the WL C_ells when baryonic feedback is active.
Here with 500:
image
Now with 100:
image
This is a little unexpected, but justifies why 500 was used in the previous CLOE. I did not test other values, perhaps 300 is enough, not sure...

cloelib/observables/photo.py
PedroCarrilho: Something like this will probably need to be done, but this changes things elsewhere substantially since we would need to find a place for the `galaxy_bias_model` class to enter. I can see a few options:
 - Change the uniformity of the C_ell calculation: i.e. we would need to use a different power spectrum for each tracer combination so we need different `get_Cl` functions (or at least different `_matter_power_spectrum_limber_grid`). 
 - Creating a new Perturbations class for each necessary case, so that its `matter_power_spectrum` method has a different P(k). This is more like in the previous version of CLOE. Probably would require a large number of different options and combinations.
 - Creating different windows. This is more like what I did, but we create several versions `get_window_positions` methods, one of which would be the real window, the others would include the necessary bias parameters. Not sure this is possible in all cases though.

cloelib/observables/photo.py
PedroCarrilho: What do you mean here by error prone? Is it that it could happen that the user passes different bias parameters that are not compatible with the necessary `galaxy_bias_model`? This could be checked, but would add extra if statements.

PedroCarrilho:
Regarding the jax stuff, here are some points that we should figure out. 
There are at least 4 places where if statements currently appear:
- In the tracer's __init__: `if 0. in z:` raising an error (I am told raising errors is not jax compatible too)
- In the above PositionsTracer `__init__` choosing the bias model. For this, I was able to figure out the jax way of doing it (I hope):
```
# Using dict.get so I can provide a default since lax has to compile every branch of the conditional
        def per_bin_case():
            bias_array = np.asarray([nuisance_params.get('b1_photo_bin%d'%bin, 1.0) for bin in range(self.n_z_bins)])
            # lax required same size for all cases, so padding here and will only use first n_z_bins values later
            return np.pad(bias_array,(0, self.z.shape[0] - self.n_z_bins))

        def per_bin_int_case():
            bias_array = np.asarray([nuisance_params.get('b1_photo_bin%d'%bin, 1.0) for bin in range(self.n_z_bins)])
            index_max_nz = np.argmax(dndz,axis=1)
            z_nz_max = jax.vmap(lambda i: lx.dynamic_index_in_dim(self.z, i, keepdims=False))(index_max_nz)
            return interpax.interp1d(self.z, z_nz_max, bias_array, extrap=True)

        def poly_case():
            poly_order = 3
            bias_array = np.asarray([nuisance_params.get('b1_photo_poly%d'%bin, 1.0) for bin in range(poly_order+1)])
            return bias_array[0] + bias_array[1] * z + bias_array[2] * z ** 2 + bias_array[3] * z ** 3

        conditions = np.array([self.flags['galaxy_bias_model'] == 'per_bin',
                               self.flags['galaxy_bias_model'] == 'per_bin_int',
                               self.flags['galaxy_bias_model'] == 'poly'])
        index = np.argwhere(conditions, size=1).squeeze()

        self.bias_array = lx.switch(index, [per_bin_case, per_bin_int_case, poly_case])
```
- the ifs in `get_window_positions`, which I can probably fix in the same way as above
- the ifs in the mixing matrix multiplication to select the tracer.

We could probably fix all of these now or in some other task, but if we want to fix them to make jax work, we would need to do it everywhere (unless I misunderstand something). Let me know if you want me to do something about it @gcanasherrera @marcobonici

cloelib/cosmology/camb_cosmology.py
```diff
@@ -327,8 +327,7 @@ def matter_power_spectrum(self, zs, ks, hubble_units=False,
             Nonlinear matter power spectrum at the specified scale
             and redshift
         """
-        pk_values = camb.get_matter_power_interpolator(
-            self.background.interface_args['CAMBparams'],
+        pk_values = self.results.get_matter_power_interpolator(
```
PedroCarrilho, line 330: I was expecting a comment here, so let me ask for your opinion instead @gcanasherrera , @marcobonici . I think this was the way I had it last time I merged, but it got changed back. This is different to the linear case, since Chiara needed the linear case to be the way it is (with camb.get...), but it is actually not correct for the nonlinear case (gives the wrong result when mead2020_feedback is activated, since that is not default), so I changed it back again. Let me know what you prefer to be done here.

marcobonici:
Regarding the three points raised by @PedroCarrilho :
- the first if is in the constructor and that part does NOT need to be differentiated, so it's fine:)
- regarding what you did, it's correct! Push it as a new commit and I'll review it (but it looks correct)
- regarding the mixing matrix: this is related to what I said to @AndreaPezzotta regarding the scale cuts. My idea is that we should create the window matrix at the very beginning of the analysis, once, and just use it later. This addresses the issue of creating a window matrix. The problem is: what about the theory vector that needs to be convolved? This should be easier to do (as it's a matter of "ordered flattening") and I have some ideas for it:)

PedroCarrilho:
I am confused @marcobonici. I thought we were expecting to have to re-instantiate every class for every point in parameter space, which would mean that when taking a derivative of the likelihood, we would have to pass through the constructor. I suppose this goes back to a discussion about how to write the likelihood, but we should figure it out soon.

Beyond this, if the constructor can have standard if statements, then what I did is not useful (don't worry though, I learned a lot), since those changes are also in the constructor. In that case I would not push anything as I think this is somewhat less readable than the version with if statements.

On the mixing matrix I agree that it is doable in various ways, I just meant we will have to do it (probably today) and not sure if we have the time

chiaramoretti:
>     * regarding the mixing matrix: this is related to what I said to @AndreaPezzotta regarding the scale cuts. My idea is that we should create the window matrix at the very beginning of the analysis, once, and just use it later. This addresses the issue of creating a window matrix. The problem is: what about the theory vector that needs to be convolved? This should be easier to do (as it's a matter of "ordered flattening") and I have some ideas for it:)

Shall we discuss the mixing matrix in a separate issue? My understanding -- at least for spectro -- is that it cannot be cut with kmax, but needs to go up to kmax+something and that something depends on many things, cannot be determined a priori

marcobonici:
Regarding what @chiaramoretti wrote: you are right! The mixing matrix can be conveniently written as a 2D matrix where one of the indices goes over the element of the window-less theoretical prediction that you are convolving. The other element, goes of the element of the data vector that you are gonna feed to your chi square. **It is the latter that need to be cut to determine the scale that you are implementing.** This is of course is analysis dependent, but we can implement a function that computes it once per each MCMC. But let us discuss in another moment.

marcobonici:
Regarding  what @PedroCarrilho wrote: sorry, it is my bad, I used a jargon that was clearly misleading. I'll try to push a notebook that uses HMC with the background to do a SNIA like analysis. In this way, things should be more clear :)

PedroCarrilho:
@marcobonici , I now added the lax conditionals to the two places where I had introduced if statements in this PR. Let me know if you think something is not correct.

marcobonici:
Hi @PedroCarrilho , I am ok with the way you implemented it.
I just need to check one thing, related to the way you are doing the calculation (without explicitely passing the dictionaries).
But we should be good to go. You laready tested it within the notebooks, right?

PedroCarrilho:
I am not sure I understood what you meant, but yes, I have checked all of this in notebooks and it all works fine.

cloelib/observables/photo.py
gcanasherrera: I think I am lining towards @marcobonici suggestion. Let's discuss in #76

marcobonici:
@PedroCarrilho , I am ok with what you wrote.
After fixing the conflict (it is the same thing I have to do with my PR), we are good to go (tagging @gcanasherrera as this is related to her PR).

gcanasherrera:
So, please guys, apologies for commiting to main. I will fix that in #129, but just put yourself up to date with main and then I will move forward with my PR. Thanks.

PR #128, opened 2025-04-04
gh-7: change of basis in jaxperturbations

### LINKED ISSUE #7: Change of basis in JaxPerturbations
The Eisenstein-Hu linear power spectrum is going be (at least for the moment) the only differentiable option for the Linear Matter Power Spectrum. However, it does not work in $A_s$ but rather in the $\sigma_8$ basis.

This requires either:
- We strip $A_s$ and include only $\sigma_8$
- We include both in the constructor
I would go with the first one, but let's discuss.

marcobonici:
The goal of this PR is to deal with the last missing piece in our JAX implementations: the perturbations.
I have slightly changed the constructor of the background: given the same previous argument, the constructor will also evaluate a bunch of derived parameters, that are required for the calculations performed afterwords.

Closes #7

### DIFF
--- a/cloelib/cosmology/jax_cosmology.py
+++ b/cloelib/cosmology/jax_cosmology.py
@@ -66,7 +66,7 @@ class JAXBackground:
         self.interface_args['JAXparams']['Omega_cdm'] = self.Omega_cdm0
         self.interface_args['JAXparams']['Omega_k'] = self.Omega_k0
         self.interface_args['JAXparams']['Omega_m0'] = self.Omega_m0
-        self.interface_args['JAXparams']['Omega_mnu0'] = self.Omega_mnu0
+        self.interface_args['JAXparams']['Omega_nu0'] = self.Omega_nu0
         self.interface_args['JAXparams']['n_s'] = self.ns
         self.interface_args['JAXparams']['m_ncdm'] = self.mnu
         self.interface_args['JAXparams']['A_s'] = self.As
--- a/cloelib/cosmology/jax_cosmology.py
+++ b/cloelib/cosmology/jax_cosmology.py
@@ -566,7 +566,7 @@ class JAXNonLinearPerturbations(Perturbations):
         a_s = a_z(zs)
 
         # Compute the linear power spectrum
-        pklin = self.linearperturbations.linear_matter_power_spectrum(ks, zs)
+        pklin = self.linearperturbations.matter_power_spectrum(zs, ks)
 
         # Compute non linear scale, effective spectral index and curvature
         k_nl, n, C = self._halofit_parameters(zs)
@@ -637,7 +637,7 @@ class JAXNonLinearPerturbations(Perturbations):
         pk_nl = 2.0 * np.pi**2 / ks**3 * d2nl
         return pk_nl.squeeze()
 
-    def nonlinear_matter_power_spectrum(self, ks, zs):
+    def matter_power_spectrum(self, ks, zs):
         """Computes the non-linear matter power spectrum.
 
         This function is just a wrapper over several nonlinear power spectra.
--- a/cloelib/cosmology/jax_cosmology.py
+++ b/cloelib/cosmology/jax_cosmology.py
@@ -460,7 +460,7 @@ class JAXLinearPerturbations:
         y = simps(int_sigma, np.log10(kmin), np.log10(kmax), N = 256)
         return 1.0 / (2.0 * np.pi**2.0) * y
 
-    def matter_power_spectrum(self, zs, ks):
+    def matter_power_spectrum(self, zs, ks,  hubble_units=False, k_hunit=False):
         r"""Computes the linear matter power spectrum.
 
         Parameters
@@ -478,7 +478,19 @@ class JAXLinearPerturbations:
             and scale factor.
 
         """
+        h = self.background.h
+
+        def k_units_case(k):
+            return k/h
+
+        def kh_units_case(k):
+            return k
+
+        conditions = np.array([k_hunit==False, k_hunit==True])
+        index = np.argwhere(conditions, size=1).squeeze()
+
         ks = np.atleast_1d(ks)
+        ks = lx.switch(index, [k_units_case, kh_units_case], ks)
         zs = np.atleast_1d(zs)
         g = self.growth_factor(zs)
         t = self.transfer_Eisenstein_Hu(ks)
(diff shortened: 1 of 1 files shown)


cloelib/cosmology/jax_cosmology.py
```diff
@@ -52,18 +52,27 @@ def __init__(self, H0: float, Omega_b0: float, Omega_cdm0: float, Omega_k0: floa
...
+                                                  self.w0, self.wa)
 
         # Initialize JaxBgk parameters
         self.interface_args = {'JAXparams': {}}  # Use a dictionary for CLASS parameters
         self.interface_args['JAXparams']['H0'] = self.H0
         self.interface_args['JAXparams']['Omega_b'] = self.Omega_b0
         self.interface_args['JAXparams']['Omega_cdm'] = self.Omega_cdm0
         self.interface_args['JAXparams']['Omega_k'] = self.Omega_k0
+        self.interface_args['JAXparams']['Omega_m0'] = self.Omega_m0
+        self.interface_args['JAXparams']['Omega_mnu0'] = self.Omega_mnu0
```
santiagocasas: this is giving me an error @marcobonici . The Omega_mnu0 doesn't exist .

cloelib/cosmology/jax_cosmology.py
```diff
@@ -190,36 +199,36 @@ def Omega_m(self, zs: np.ndarray) -> np.ndarray:
...
+        Args:
+            background (Background): A Background instance.
+            redshifts (np.ndarray): Array of redshifts for the calculations.
         """
         self.background = background
+        self.z = redshifts
 
     def w_a(self, a):
-        return self.background.w + (1.0 - a) * self.background.wa  # Equation (6) in Linder (2003)
+        return self.background.w0 + (1.0 - a) * self.background.wa  # Equation (6) in Linder (2003)
```
santiagocasas: this is literally a background quantity, so I would move it there and just use self.background.w_a()

cloelib/cosmology/jax_cosmology.py
```diff
@@ -190,36 +199,36 @@ def Omega_m(self, zs: np.ndarray) -> np.ndarray:
...
     def f_de(self, a):
-        return -3.0 * (1.0 + self.background.w + self.background.wa) * np.log(a) + 3.0 * self.background.wa * (a - 1.0)
+        return -3.0 * (1.0 + self.background.w0 + self.background.wa) * np.log(a) + 3.0 * self.background.wa * (a - 1.0)
 
     def Esqr(self, a):
-        Omm = self.background.Omb + self.background.Omc
-        OmDE = 1. - Omm - self.background.Omk
-        return (Omm * np.power(a, -3) + self.background.Omk * np.power(a, -2)
+        OmDE = 1. - self.background.Omega_m0 - self.background.Omega_k0
+        return (self.background.Omega_m0 * np.power(a, -3) + self.background.Omega_k0 * np.power(a, -2)
```
santiagocasas: same for this function (which is E(z)^2) and its associated f_de.  You already have this in Background and repeating code is very dangerous. On top of the fact that this is confusing for people developing Background and Perturbations. I would move it to Background.

cloelib/cosmology/jax_cosmology.py
```diff
@@ -190,36 +199,36 @@ def Omega_m(self, zs: np.ndarray) -> np.ndarray:
...
                 + OmDE * np.exp(self.f_de(a)))
 
     def Omega_m_a(self, a):
-        Omm = self.background.Omb + self.background.Omc
-        return Omm * np.power(a, -3) / self.Esqr(a)
+        return self.background.Omega_m0 * np.power(a, -3) / self.Esqr(a)
 
     def Omega_de_a(self, a):
-        OmDE = 1. - self.background.Omb - self.background.Omc - self.background.Omk
+        OmDE = 1. - self.background.Omega_m0 - self.background.Omega_k0
```
santiagocasas: And same for these two, Omega_m and Omega_de.  You probably should just define an a(z) and z(a) function and use the background quantities.

cloelib/cosmology/jax_cosmology.py
```diff
@@ -851,3 +855,51 @@ def Pkl_interp(k_l, z_l, ks, zs, Pk):
                                  method="cubic")
 
 Pkl_interp_vmap = jax.jit(jax.vmap(Pkl_interp, in_axes=(0, None, None, None, None)))
+
+#from 2410.14623
+def As_to_sigma8_max_precision(As, Om, Ob, h, ns, mnu, w0, wa):
```
santiagocasas, line 881: this is great!  I would add all these functions to some kind of jax_utils.py or maybe another class on its own in this same module. But for now and to finish this quickly, it is fine if all of this is just here.

santiagocasas (changes requested):
Thanks for the work @marcobonici !  Everything looks great, except a typo in a variable and moving the background things to background!

cloelib/cosmology/jax_cosmology.py
marcobonici: I'll correct the typo.

cloelib/cosmology/jax_cosmology.py
marcobonici: You are right, I was lazy. I will add all of these methods to the background!

cloelib/cosmology/jax_cosmology.py
marcobonici: If you ok, for the sake of being fast, I will keep this as it is and open a new issue (that I will likely start tackling right after this PR is merged😂)

cloelib/cosmology/jax_cosmology.py
marcobonici: Typo fixed.

cloelib/cosmology/jax_cosmology.py
marcobonici: Moved to the background.

marcobonici:
Hi @santiagocasas , we should be there:
- I solved the issue with the typo
- I moved the background related functions to the right place

The only things I did not do are:
- Changing the argument of the additional quantities, from $a$ to $z$. This would required rewriting the ODE related to the growth factor integration (and I am too short on sleep to do that now).
- Moving some of the functions into the utils
If you ok with this, I will open two issue related to those and will start working on them after this PR is merged

santiagocasas:
Thanks @marcobonici !!  
Yeah, there is no need to actually change the arguments of the Omega_m_a for instance. You keep it as it is, and inside you transform a to z and call the other function you had before. Same for E(a). 
But for the sake of speed, let's keep that for later!


I will test quickly and merge!

marcobonici:
Hi @santiagocasas 
I fixed it
image
The residuals mistakes were:
- The util to convert from As to sigma8 was missing to multiply As by 10^9, as this is the quantity required by the function
- Previously, the matter power spectrum was getting as input `ks` and `zs` and the order has now been changes (but no updated in a couple of points

Let me know what you think, but we should be almost ready. I just have to add a small piece (possibility to convert to Hubble units!) and we should be good to go.

santiagocasas:
thanks @marcobonici !!  reviewing this now!

marcobonici:
@santiagocasas I fixed the issue for the mergeability :)

santiagocasas:
@marcobonici I am merging now, correcting the merge conflict!

PR #141, opened 2025-04-16
gh-140: Added function to compute the Legendre multipoles of a specific diagr…

### LINKED ISSUE #140: Add function to obtain Legendre multipoles of individual diagrams (for analytical marginalisation)
## 📝 Issue Summary
So far we have a single function to obtain the total Legendre multipoles, corresponding to all possible one-loop coupling of density and velocity field, plus counterterms and shot-noise. To implement analytical marginalisation, we need to do the same with the indivual terms entering at linear level in the model.

- [ ] **Goal 1**: Implement function to obtain multipoles of individual terms
- [ ] **Goal 2**: Split the shot-noise in legendre_multipoles into its three components, to allow also shot-noise to be processed by the function created at step 1.

### ✅ Expected Outcome
It will be possible to obtain Legendre multipoles of the individual diagrams of the loop expansion in a similar way as what we currently do for the total multipoles

### 👥 Assignees
- @AndreaPezzotta

AndreaPezzotta:
### ✅ Summary
Added function power_X_multipoles to obtain the Legendre multipoles of a specific PT diagram (loop, counterterms, noise). Related to the noise terms, these have been defined individually and not with a single noise function, to allow for analytical marginalisation of any of these terms on their own. Also modified the interface functions in the comet spectro power classes (EFT and VDG) to accept a list of names for the specific diagrams different from the native ones of comet.

### 🔄 Changes
- [ ] Added power_X_multipoles function in legendre_multipoles.py (same structure of power_multipoles)
- [ ] Split 2d noise function into the separate contributions noise_k0, noise_k2, and noise_k2mu2
- [ ] Changed interface functions in the comet spectro power classes to adapt them to the naming convention of cloelib (the latter can be changed eventually)

### 📝 Documentation
- [ ] Added documentation for the new functions

### 🏗 Related Issues
Resolves #140
---

- [ ] I have titled this PR before merging as "gh-#:", where "#" represents the task it closes
- [ ] My code follows the repository's coding style
- [ ] I have tested my changes locally
- [ ] No new warnings or errors introduced
- [ ] My changes do not introduce breaking changes

- [x] Check that there are no `No newline at the end of file` warnings
- [x] Check that the implementation follows the contributing guidelines and style choices

### DIFF
--- a/cloelib/observables/CometEFT_spectro.py
+++ b/cloelib/observables/CometEFT_spectro.py
@@ -82,31 +82,31 @@ class CometEFT_SpectroPower:
             comet_inst.P2d_nostoch(k=k[:, :, np.newaxis], mu=mu[:, np.newaxis],
                                    params=self.parameters, de_model='w0wa'))
 
-    def Pk2d_X_rsd(self, k: np.ndarray, mu: np.ndarray, X_list: list) -> np.ndarray:
-        r"""2D power spectrum for the specific diagram X of the loop expansion
+    def Pk2d_term_rsd(self, k: np.ndarray, mu: np.ndarray, term_list: list) -> np.ndarray:
+        r"""2D power spectrum for a subset of specific diagrams of the loop expansion
         Parameters
         ----------
         k: np.ndarray
             Wavenumber
         mu: np.ndarray
             Angle (cosinus) to the line of sight
-        X: list
+        term_list: list
             Identifiers of loop diagrams
         Returns
         -------
-        PX2d_rsd: np.ndarray
-            2D power spectrum of term X
+        Pk2d_term_rsd: np.ndarray
+            2D power spectrum of specific terms
         """
-        X_list_expanded, index_map = [], {}
-        for key in X_list:
+        term_list_expanded, index_map = [], {}
+        for key in term_list:
             vals = self.diagram_naming_relation[key]
             vals = vals if isinstance(vals, list) else [vals]
-            index_map[key] = list(range(len(X_list_expanded),
-                                  len(X_list_expanded) + len(vals)))
-            X_list_expanded.extend(vals)
+            index_map[key] = list(range(len(term_list_expanded),
+                                  len(term_list_expanded) + len(vals)))
+            term_list_expanded.extend(vals)
         Pk2d_expanded = np.squeeze(
             comet_inst.PX_2d(k=k[:, :, np.newaxis], mu=mu[:, np.newaxis],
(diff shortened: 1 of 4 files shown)


chiaramoretti:
Hi @AndreaPezzotta I have gone through your implementation, left a few comments here and there, nothing major.
Two more general comments:
* I think using `X` might be confusing, could we change this to either `term` or `marg_term` (or similar)?
* Do we want this to be a general thing, i.e. multipoles for any single term can be obtained, or do we want to allow it only for linear/marginalised parameters? For the first scenario everything's good, but we need to be careful in the likelihood implementation to only allow for linear parameters (and I would go with `term` for the previous point). For the second case we need to add a check or something to only allow for the linear parameters to be computed by `Pk2d_X_rsd` and `power_X_multipoles`. A third option would be to allow for any term in the `SpectroPower` implementation but only allow for the linear parameters in `LegendreMultipoles`

Final thought, also related to the second point above, is that the implementation with the diagram names might be an issue for other external codes, as they might not output the single terms (e.g. for PBJ the Pk2d method I implemented outputs the sum already, I can change it but not sure it will be possible/easy for other codes). Not super-important at this point so we might want to revisit later, if needed.

AndreaPezzotta:
Thanks @chiaramoretti! I have pushed a very minor commit with the modification `X` -> `term`. Hopefully I have spotted all occurrences, but the notebook in playground works locally so it should be good.

About the other points that you were making, I believe there should be in principle no harm in leaving the possibility of evaluating the complete set of diagrams. Clearly it is up to us to only call the specific subset that can be analytically marginalised within the likelihood. But in terms of testing it could be useful to leave also the other terms. Unless there is something that I have not thought about that may be potentially dangerous.. About asking for the individual list of terms or for the combined one, I believe you are meaning to have a combined function that returns both the non-marginalisable Pk2d and the individual marginalisable diagrams, as opposed to have two separate functions as we are currently doing here (?) If so, maybe there are ways of leaving a bit more of freedom to the individual codes, but not sure how much this will be possible with the current protocol structure (maybe mixings are the way, but I am no expert).

For this specific thing, though, I would say that it should be relatively easy for the nonlinear codes to adapt to a common structure, in case it needs be. Consider that we should  anyway set a minimal interface that the external codes must satisfy. And for comet I also had to change a few things internally to make it work with the current interface. Not saying that this is going to be the final one, jut pointing out that most likely external codes will have to be slightly massaged by their developers if we want to plug them into cloelib. But happy to discuss it further

chiaramoretti:
All good, merging!

PR #214, opened 2025-06-19
gh-213: correlation function speedup

### LINKED ISSUE #213: Correlation function speed-up
## 📝 Issue Summary
In this moment the computation of the angular correlation function, based on Wigner d-function computation, is the bottle neck of the correlation function computation, leading to a computing time of 20 seconds per function. Improve this, by implementing a jax-compatible cache mechanism.

- [ ] **Goal 1**: Speedup the Wigner matrices computation by using a cache mechanism.

### ✅ Expected Outcome
When the issue is solver, we expect the function to be actually faster than it was before.

### 👥 Assignees
- @marcobonici

marcobonici:
### ✅ Summary
The goal of this PR is to implement a cache mechanism such that computation of the wigner matrices does not impact the performance of the angular correlation function evaluation.

### 🔄 Changes

- [ ] Implements a cache mechanism for Wigner function computation

### 🛠 How to Test
Checked that the results do not change, as the goal of this is not to change the output of a computation, but to avoid repeated computations.

### 📝 Documentation

### 🏗 Related Issues

Fixes #213 

Fun fact: previously the comparison with `ccl` was showing
image
image

But this is when we are using the *Bessel* method from `ccl`, which means when we are considering a flat sky. But we are doing this in the full sky, as we are using the Wigner matrices!
So, if we turn on the *Legendre* method in `ccl` (e.g. we switch to full sky) we get

image
image

(If you use the keyword argument "Bessel" in the validation notebook you will get the old results.)
With the full sky calculation, the results are in closer agreement (I think that at some point @PedroCarrilho was slightly concerned about the accuracy of our calculations, but now everything should provably be good :) ).

---

- [x] My code follows the repository's coding style
- [x] My changes do not introduce breaking changes

- [x] Check that there are no `No newline at the end of file` warnings
- [x] Check that the implementation follows the contributing guidelines and style choices
- [x] Check that the corresponding branch has been deleted after merging. If not, delete it


cloelib/summary_statistics/angular_correlation_function_wigner.py
```diff
@@ -12,15 +12,49 @@
...
 import jax
 from jax import jit
-
-#from jax.scipy.special import gammaln
+from functools import lru_cache, wraps
 
 from cloelib.observables.photo import ShearTracer #, PositionsTracer
 from .angular_correlation_function import AngularCorrelationFunction
 
+def memoize_jax(func):
```
gcanasherrera: I wonder if this is the best place to do the cache. What if we try to homogenise with the angular_two_point side?

cloelib/summary_statistics/angular_correlation_function_wigner.py
```diff
@@ -12,15 +12,49 @@
...
+        }
+        return cached_func(*hashable_args, **hashable_kwargs_dict)
+
+    return wrapper
 
 @jit
-def d_0_0_ell(beta, ell):
+def _d_0_0_ell_compute(beta, ell):
     """
     Compute d_00^ell(beta).
```
gcanasherrera: I know it's not part of the code task, but maybe add the arguments into the API docs

cloelib/summary_statistics/angular_correlation_function_wigner.py
```diff
@@ -180,11 +214,20 @@ def compute_d_2_0(ell):
 # Vectorized versions of Wigner d-matrix functions
 # -----------------------------------------------------------------------------------
 
-d_0_0_vmap = jax.vmap(jax.vmap(d_0_0_ell, (None, 0)), (0, None))
-d_2_2_vmap = jax.vmap(jax.vmap(d_2_2_ell, (None, 0)), (0, None))
-d_2_m2_vmap = jax.vmap(jax.vmap(d_2_m2_ell, (None, 0)), (0, None))
+d_0_0_ell = memoize_jax(_d_0_0_ell_compute)
```
gcanasherrera, line 257: I would personally leave some comments here for the record

gcanasherrera (commented):
hi @marcobonici , this looks great. I left some comments but I have some suggestions, if you don't mind:

- Could you add unit tests that explicitly check caching behavior? I know we don't have them in the Cells, but we will need there too. Better to get started somewhere.
- I am concerned a bit the memory usage of lru_cache in large-computations (like MCMCs), just to make sure it does not cause excessive memory growth (because we will have double cache computation, from Cells --> Xis). Could you please run a for-loop and check a bit?

I know I am being picky but with exceptional developers that's more like a duty :)

marcobonici:
Hi @gcanasherrera ,
regarding the first two comments:
1. What do you mean by "caching behavior"? You mean to compare the values retrieved by the cached and not cached functions?
2. The cache size here should not be a problem, but I will do that :)

gcanasherrera:
1. What I mean is to have a unit tests that indeed checks that the speed improves due to caching, or that, indeed, the catching works. Nothing in terms of comparing with another computation.
2. Thanks!

Also, other comments above.

gcanasherrera:
@marcobonici any news I've missed about this?

marcobonici:
Hi @gcanasherrera ,
1. I have added tests that ensure that the cached and the non cached version retrieve the same numbers.
2. I checked locally, with a for loop, that there are no memory issues. I called the cached function 100'000 times and there was no issue.

gcanasherrera:
Thanks for this @marcobonici , almost there, but you need to address my comments of the review itself, please :) 

Appreciated! I know you're running busy these last days.

cloelib/summary_statistics/angular_correlation_function_wigner.py
marcobonici: Most likely, it isn't. I movied it into a dedicated module for caching in auxiliary (as in this manner should be more easily accessible for other modules).

cloelib/summary_statistics/angular_correlation_function_wigner.py
marcobonici: I have added a few comments.

cloelib/summary_statistics/angular_correlation_function_wigner.py
marcobonici: I have modified the docstrings. Now they include more information.

marcobonici:
Hi @gcanasherrera , sorry for the delay.

I have addressed all the comments related to the PR and checked that the unit tests are passing.

cloelib/auxiliary/cache.py
```diff
@@ -0,0 +1,57 @@
...
+Caching utilities for functions that accept JAX arrays.
+
+This module exposes `memoize_jax`, a decorator that adds **hash-based
+memoization** to any Python function whose positional or keyword
+arguments include JAX ``DeviceArray`` objects.
+
+Background
+----------
+JAX arrays are mutable and therefore *unhashable*, so they cannot be
+passed directly to ``functools.lru_cache``[19].  `memoize_jax` solves
```
gcanasherrera: which are these [19] and [16]?

cloelib/summary_statistics/angular_correlation_function_wigner.py
```diff
@@ -133,12 +185,33 @@ def compute_d_2_2(ell):
...
+    \(d^{\ell-1}_{20}\) and \(d^{\ell-2}_{20}\).
+    Beyond a configurable threshold (ℓ ≈ 30 000) the stable three-term
+    *approximation*
+    \(d^{\ell}_{20} ≈ 2\,d^{1}_{00}\,d^{\ell-1}_{20}-d^{\ell-2}_{20}\)
+    is used to avoid numerical overflow while remaining compatible with
+    `jax.jit`.
+
+    Parameters
+    ----------
+    beta : float or jax.numpy.ndarray
```
gcanasherrera: I know I'm being a pain, but I also need to learn to be a better reviewer to keep the standards high. If you are specifying what they should be in the docs, you could use here types to make sure that only accept those types.

cloelib/summary_statistics/angular_correlation_function_wigner.py
```diff
@@ -13,40 +13,73 @@
 import jax
 from jax import jit
 
-#from jax.scipy.special import gammaln
-
 from cloelib.observables.photo import ShearTracer #, PositionsTracer
 from .angular_correlation_function import AngularCorrelationFunction
-
+from cloelib.auxiliary.cache import memoize_jax
```
gcanasherrera, line 19: This looks way more neat! Now we need to make sure that the shear kernel gets this neat version too. Could you open a follow up task?

gcanasherrera (commented):
Last minimum questions.

cloelib/summary_statistics/angular_correlation_function_wigner.py
marcobonici: I am opening it!

cloelib/auxiliary/cache.py
marcobonici: Removed them , they were useless.

cloelib/summary_statistics/angular_correlation_function_wigner.py
marcobonici: Added.

cloelib/summary_statistics/angular_correlation_function_wigner.py
marcobonici: Opened.

marcobonici:
Hi @gcanasherrera , I have implemented the last few things. Ready to go!

gcanasherrera (approved):

gcanasherrera:
Great Marco, merging now!

PR #161, opened 2025-05-19
gh-57:  cosmology from bao

chiaramoretti:
### ✅ Summary
Implements computation of alphas to get cosmology from BAO measurements

### 🔄 Changes
* Added `bao_alphas.py` in `summary_statistics`, implements `BAO_alphas` class. `BAO_alphas` requires definition of a background and background_fiducial, as well as a list of redshifts 
* the class implements separate methods for `alpha_par`, `alpha_perp`, `alpha_iso` and `alpha_AP`, as well as a method that calls all these and puts results in a dictionary for easy access
* also implements a `sound_horizon_drag` method

### 🛠 How to Test
See [this branch of playground](https://github.com/cloe-org/playground/tree/31-add-bao-notebook) to see an example notebbok

### 📝 Documentation
- [ ] This PR updates documentation
- [ ] This PR does not require documentation changes
- [ ] Issue created for documentation update: Don't forget to link the task!

### 🏗 Related Issues
Resolves #57 

### 📌 Additional Notes
Please see potential issues listed in #57 

---

### ✅ PR Checklist for Developers
- [x] I have titled this PR before merging as "gh-#:", where "#" represents the task it closes
- [x] My code follows the repository's coding style
- [x] I have tested my changes locally
- [x] No new warnings or errors introduced
- [x] I have updated documentation (if applicable)
- [x] My changes do not introduce breaking changes
- [ ] I have added tests (if applicable)
- [x] I have consistently updated the GitHub information for the project, including milestones, task types, and other relevant details.

### ✅ PR Checklist for Reviewers
- [ ] The next PR targets the correct branch
- [ ] CI tests have run and passed for the latest commit on the source branch
- [ ] Check that the code can still be installed if new packages are imported
- [ ] If necessary, the notebooks in the `playground` will be updated in a corresponding follow-up PR
- [ ] Coverage percentage is retained or increased
- [ ] Quality of new/changed code is acceptable
- [ ] Quality of new/changed unit tests is acceptable
- [ ] No data files have been included in the commits
- [ ] Implementation follows the agreed task description point by point
- [ ] Check that there are no `No newline at the end of file` warnings
- [ ] Check that any added folder/file has been added to the `README.md` file
- [ ] Check that the implementation follows the contributing guidelines and style choices
- [ ] Check that the documentation has been updated accordantly
- [ ] Check that the corresponding branch has been deleted after merging. If not, delete it

PR #221, opened 2025-06-20
gh-220: implementation of Gauss-Legendre integration for power spectrum multipoles

### LINKED ISSUE #220: Update integration method to obtain Legendre multipoles
## 📝 Issue Summary
Right now we are using Simpson's rule on a fixed mu array of length 101. This can be improved by adopting a Gauss-Legendre quadrature routine, to make the integration faster while keeping the same level of accuracy.

Additionally, this task is meant to fix a bug introduced in one of the latest PRs. This is related to the definition of the AP normalisation factor in the routine that computes the Legendre multipoles of the individual loop corrections (see #221 for more details)

- [ ] **Goal 1**: Substitute Simpson's rule with Gauss-Legendre
- [ ] **Goal 2**: Fix bug

### ✅ Expected Outcome
The integration of the angular variable to obtain Legendre multipoles is going to be faster

### 📌 Additional Notes
Here's an example of the performances between two set of multipoles, one obtained with SImpson's rule and one with the Gauss-Legendre method:

Image

and the time performances after the first evaluation are of the order of 40ms using the current integration routine and 10ms using Gauss-Legendre (these numbers are estimated on a laptop with 10 threads)

### 👥 Assignees
- @AndreaPezzotta

AndreaPezzotta:
### ✅ Summary
This PR implements a new way of intregrating the anisotropic P(k,mu) along the angular variable mu. In the current version of the code, we are using Simpson's rule. With this PR we are switching to a more efficient Gauss-Legendre quadrature routine. This preserves the same level of accuracy of Simpson's rule (as shown in the description of #220) but achieves 3-4 times less time to evaluate.

Additionally, during the implementation for this task, I spotted a bug introduced during one of the last PR. Instead of computing the AP normalisation factor for the amplitude of the power spectrum multipoles as $`q_\perp^2 q_\parallel`$, the function to obtain the multipoles of the individual loop contributions was erroneously returning $`q_\parallel^3`$

### 🔄 Changes

- [ ] Implemented Gauss-Legendre quadrature routine
- [ ] Fixed bug related to AP parameters

### 📝 Documentation

### 🏗 Related Issues
Resolves #220

---

- [ ] I have titled this PR before merging as "gh-#:", where "#" represents the task it closes
- [ ] My code follows the repository's coding style
- [ ] I have tested my changes locally
- [ ] No new warnings or errors introduced
- [ ] My changes do not introduce breaking changes

### DIFF
--- a/cloelib/summary_statistics/legendre_multipoles.py
+++ b/cloelib/summary_statistics/legendre_multipoles.py
@@ -41,10 +41,8 @@ class LegendreMultipoles:
         self.background_fiducial = background_fiducial
         self.ap_distortion = APDistortion(spectro_power.background, background_fiducial)
 
-        mu_min = 0.0
-        mu_max = 1.0
-        mu_samp = 101
-        self.mu_grid = np.linspace(mu_min, mu_max, mu_samp)
+        self.mu_grid, self.mu_weights = np.polynomial.legendre.leggauss(10)
+        self.mu_grid = 0.5 * (self.mu_grid + 1.0)
 
         self.parameters = parameters
         self.nbar = nbar
@@ -233,18 +231,15 @@ class LegendreMultipoles:
                      self.ap_distortion.q_AP_lo(self.redshift) if use_AP else 1.0)
         prefactors = np.array([(2.0 * m + 1.0) for m in ells]) / 2.0 / \
             AP_factor
+        kAP = self._k_AP(k, self.mu_grid, self.redshift, use_AP=use_AP)
+        muAP = self._mu_AP(self.mu_grid, self.redshift, use_AP=use_AP)
+        Pk2d_tot = self._Pk2d_tot(kAP, muAP)
         multipoles = {}
         for i,ell in enumerate(ells):
+            leg = legendre(ell, self.mu_grid)
             multipoles[f'ell{ell}'] = \
-                integrate.simps(self._Pk2d_tot(self._k_AP(k, self.mu_grid,
-                                                          self.redshift,
-                                                          use_AP=use_AP),
-                                               self._mu_AP(self.mu_grid,
-                                                           self.redshift,
-                                                           use_AP=use_AP)) *
-                                legendre(ell, self.mu_grid),
-                                self.mu_grid, axis=1)
-            multipoles[f'ell{ell}'] *= (2.0 * prefactors[i])
+                np.einsum("ab,b,b->a", Pk2d_tot, leg, self.mu_weights)
+            multipoles[f'ell{ell}'] *= prefactors[i]
         return multipoles
 
     def power_term_multipoles(self, k: np.ndarray, term_list: list,
@@ -269,7 +264,7 @@ class LegendreMultipoles:
         """
         ells = self._ensure_array(ells) \
             if ells is not None else np.array([0,2,4])
-        AP_factor = (self.ap_distortion.q_AP_lo(self.redshift)**2 *
+        AP_factor = (self.ap_distortion.q_AP_tr(self.redshift)**2 *
                      self.ap_distortion.q_AP_lo(self.redshift) if use_AP else 1.0)
         prefactors = np.array([(2.0 * m + 1.0) for m in ells]) / 2.0 / \
             AP_factor
@@ -293,10 +288,10 @@ class LegendreMultipoles:
                 else noise_func[term_list[index]](kAP) for index in noise_ids])
         multipoles = {}
         for i,ell in enumerate(ells):
+            leg = legendre(ell, self.mu_grid)
             multipoles[f'ell{ell}'] = \
-                integrate.simps(Pk2d * legendre(ell, self.mu_grid),
-                                self.mu_grid, axis=-1)
-            multipoles[f'ell{ell}'] *= (2.0 * prefactors[i])
+                np.einsum("abc,c,c->ab", Pk2d, leg, self.mu_weights)
+            multipoles[f'ell{ell}'] *= prefactors[i]
         return multipoles
 
     def convolved_power_multipoles(self, mixing_matrix: dict,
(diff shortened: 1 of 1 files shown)


cloelib/summary_statistics/legendre_multipoles.py
```diff
@@ -41,10 +41,8 @@ def __init__(self, spectro_power: SpectroPower,
         self.background_fiducial = background_fiducial
         self.ap_distortion = APDistortion(spectro_power.background, background_fiducial)
 
-        mu_min = 0.0
-        mu_max = 1.0
-        mu_samp = 101
-        self.mu_grid = np.linspace(mu_min, mu_max, mu_samp)
+        self.mu_grid, self.mu_weights = np.polynomial.legendre.leggauss(10)
```
santiagocasas, line 44: @AndreaPezzotta did you check if these amount of weights is necessary? Already N=4 yields errors of 10^-7, so this might be an overkill and we could save some time there.
Also (for the mid-far future) it might be good to start thinking about a `precision_setting` structure that we can modify once in the whole code and it increases the precision of the whole code consistently, instead of having these settings hard-coded and spread throughout several files.

cloelib/summary_statistics/legendre_multipoles.py
```diff
@@ -269,7 +264,7 @@ def power_term_multipoles(self, k: np.ndarray, term_list: list,
         """
         ells = self._ensure_array(ells) \
             if ells is not None else np.array([0,2,4])
-        AP_factor = (self.ap_distortion.q_AP_lo(self.redshift)**2 *
+        AP_factor = (self.ap_distortion.q_AP_tr(self.redshift)**2 *
```
santiagocasas, line 282: does this mean that there was also a bug in the previous code, since it was calling twice the `q_AP_lo` ?

cloelib/summary_statistics/legendre_multipoles.py
```diff
@@ -293,10 +288,10 @@ def power_term_multipoles(self, k: np.ndarray, term_list: list,
                 else noise_func[term_list[index]](kAP) for index in noise_ids])
         multipoles = {}
         for i,ell in enumerate(ells):
+            leg = legendre(ell, self.mu_grid)
             multipoles[f'ell{ell}'] = \
-                integrate.simps(Pk2d * legendre(ell, self.mu_grid),
-                                self.mu_grid, axis=-1)
-            multipoles[f'ell{ell}'] *= (2.0 * prefactors[i])
+                np.einsum("abc,c,c->ab", Pk2d, leg, self.mu_weights)
+            multipoles[f'ell{ell}'] *= prefactors[i]
```
santiagocasas, line 309: Just a question:  
the `2.0` factor in front of the `prefactors[i]` is now gone because the integration over mu changed to double its previous range?

santiagocasas:
@AndreaPezzotta don't forget to pull before finishing the review, I have added dosctrings to the only remaining method that had no docstrings in this module:  `_ensure_array`

santiagocasas (commented):
Hi @AndreaPezzotta here my comments on the code, had forgotten as well to hit submit review! thanks!

cloelib/summary_statistics/legendre_multipoles.py
AndreaPezzotta: I have checked the convergence as a function of few values of N, and it seems that with N=10 we are still within the convergence limit, especially considering the errorbars of the higher-order multipoles, which are the most sensitive ones to the accuracy of the integration. On the contrary, with N=4 I am getting almost a 10% offset with respect to the reference case (with N=20). So I think N=10 could be a good compromise in the end. And I agree, at some point we may want to have a module to allow the user to tweak different precision settings

cloelib/summary_statistics/legendre_multipoles.py
AndreaPezzotta: Actually this is due to an improper rescaling of the integration range. Natively this is [-1,1] but we can exploit the parity of the function to integrate only over the range [0,1]. This is why `self.mu_grid` is rescaled in that way. However to mkae the integration consistent we also need to rescale the weights, which was not done in the version that I originally committed, therefore leading to ane xtra factor of 0.5 that was cancelling out with the 2 in the line that you have referenced. So, the outcome is correct, but for clarity  I have pushed a new version where I rescale the GL weights accordingly and include the factor 2 in the integral. Let me know if this is fine!

AndreaPezzotta:
Many thanks for the review @santiagocasas ! I have pushed a new commit and replied to your comments. Let me know what you think about it

cloelib/summary_statistics/legendre_multipoles.py
santiagocasas: okay this makes sense, if your reference case is N=20. Probably, when we had implemented this in the past, our reference case was just a less accurate method like a trapezoid and it was enough with N=4 for the given purposes. Yes, let's keep in mind all the hardcoded precision settings.

cloelib/summary_statistics/legendre_multipoles.py
santiagocasas: perfect, that's much clearer now! thanks!

cloelib/summary_statistics/legendre_multipoles.py
santiagocasas: just to be completely sure here @AndreaPezzotta , now it is correct, right? before it was q_AP_lo**2 * q_AP_lo .

santiagocasas (approved):
thanks for the PR @AndreaPezzotta !  it looks all good now, I just left another comment for clarity!

cloelib/summary_statistics/legendre_multipoles.py
AndreaPezzotta: Yes, the current code on the main branch has this bug, as this was introduced in a recent PR (#161)

AndreaPezzotta:
Apologise @santiagocasas! I didn't realise I didn't press the comment button! Yes, before it was q_lo**2 * q_lo, which is wrong, as we need two powers of q_tr and one of q_lo

santiagocasas:
merging now, thanks @AndreaPezzotta !

PR #257, opened 2025-08-18
gh-215: Allow Baryon Boosts to attach to Perturbations as Mixins

### LINKED ISSUE #76: Modify Background + Perturbations + Tracers for mixins

### LINKED ISSUE #215: Include FLAMINGO Baryonresponse emulator
## 📝 Issue Summary
The FlamingoBaryonResponseEmulator for the correction of the matter power spectrum suppression is not yet implemented.

- [ ] **Goal 1**: Add emulator to /cloelib/cosmology and to files calling baryonic corrections

### ✅ Expected Outcome
FLAMINGO emulator should be able to be called when doing a cosmological inference

### 📌 Additional Notes
Flamingo emulator: https://github.com/FLAMINGOSIM/FlamingoBaryonResponseEmulator

### 👥 Assignees
- @TBA

JegerBroxterman:
### ✅ Summary

Introduces a composable baryonic boost framework in `cloelib` built around a `BaryonBoostMixin` Protocol. Three ready-to-use backends are provided (FLAMINGO, BACCOemu, HMcode2020emu), each of which produces a `Perturbations`-compatible object whose `matter_power_spectrum` already includes the baryonic correction and can be passed directly to any `cloelib` tracer or likelihood.

Additionally, all `matter_power_spectrum` methods now return a 1-D array when called with a single redshift (via `.squeeze()`), making the API consistent with scalar-redshift use cases.

### 🔄 Changes

**`cloelib/cosmology/cosmology.py`**
- Added `BaryonBoostMixin(Protocol)` with `@runtime_checkable`, defining the `baryonic_suppression(zs, ks, k_hunit=False) -> np.ndarray` interface
- Added `with_baryon_boost(NonLinearClass, BaryonMixinClass)` factory: zero-boilerplate composition of any nonlinear perturbations class with any baryonic mixin; mixin args are passed via `baryon_kwargs=dict(...)`

**`cloelib/cosmology/FlamingoBaryonResponseEmulator_cosmology.py`** *(new file)*
- `FlamingoBaryonBoostMixin`: standalone mixin implementing `baryonic_suppression` via the FLAMINGO response emulator. Parameters: `fgas_sigma`, `Mstar_sigma`, `jet_fraction`. Valid for $z \leq 3$ and $k \leq 10^{1.5}\,h/\text{Mpc}$; response is clamped/set to 1 outside those ranges
- `CAMBNonLinearFLAMINGOPerturbations`: combines `CAMBNonLinearPerturbations` + `FlamingoBaryonBoostMixin`

**`cloelib/cosmology/baccoemu_cosmology.py`**
- `BACCOemuBaryonBoostMixin`: standalone mixin that pre-computes $B(k,z) = P_\text{baryon}/P_\text{DMO}$ as a bivariate spline at init (runs emulator twice). Supports the 7-parameter Aricò+2021 baryonification model
- `BACCOemuNonLinearBaryonicPerturbations`: combines `BACCOemuNonLinearPerturbations` + `BACCOemuBaryonBoostMixin`

**`cloelib/cosmology/HMcode2020Emu_cosmology.py`**
- `HMcode2020BaryonBoostMixin`: standalone mixin that pre-computes the baryon ratio spline at init via `log10TAGN`. Scales below the nonlinear emulator $k$ range return suppression = 1
- `HMemuNonLinearBaryonicPerturbations`: combines `HMemuNonLinearPerturbations` + `HMcode2020BaryonBoostMixin`

**All `matter_power_spectrum` methods (CAMB, FLAMINGO, BACCOemu, HMcode, `with_baryon_boost` factory)**
- Return value now always has shape `(n_k,)` for single-redshift calls via `.squeeze()`. Multi-redshift calls retain shape `(n_z, n_k)`

**`README.md`**
- Updated to include dependency on `FlamingoBaryonResponseEmulator`

### 🛠 How to Test

An example of how to use the code is provided in a pull request to `cloe-org/playground`, see https://github.com/cloe-org/playground/pull/67.

The tutorial notebook (`tutorials/cosmology/baryon_boost.ipynb`) covers:
- Instantiating all three backends
- Comparing $B(k,z)$ across backends
- Computing the full 3×2pt $C_\ell$s with baryonic suppression
- DIY composition patterns using `BaryonBoostMixin` directly or via `with_baryon_boost`

### 📝 Documentation

- `README.md` updated to describe the dependency on `FlamingoBaryonResponseEmulator`
- All mixin and combined classes include docstrings documenting the baryonic parameter ranges and out-of-range behaviour
- For FLAMINGO specifically: $z > 3$ → response fixed to 1; $k > 10^{1.5}\,h/\text{Mpc}$ → response clamped to its value at the maximum trained wavenumber. This is documented in the relevant docstrings. No additional documentation location is proposed at this time, but it could be added to the main docs if desired

### 🏗 Related Issues

Resolves #215 #76

---

- [x] My code follows the repository's coding style
- [x] My changes do not introduce breaking changes
- [x] I have added tests (if applicable)


gcanasherrera:
@all-contributors add @JegerBroxterman for bug, code

MatthieuSchaller:
Hi guys,

Is there anything more we can help with here? Anything on the emulator or something like that?

gcanasherrera:
@all-contributors please add @JegerBroxterman for code, ideas

gcanasherrera:
@all-contributors please add @MatthieuSchaller for ideas

gcanasherrera:
@PedroCarrilho I've finally found time to give this a push, could you check? In particular, scientifically, that I have not messed things up with extrapolation ranges

@ntessore I know that your to-do list is very long, but, if you have only 15 minutes, they would be appreciated here to make sure I did not messed up with the protocols/inheritance. In the future I need to find a better way to have just an interface with observables through a common `matter_power_spectrum()` method only.

cloelib/cosmology/baccoemu_cosmology.py
```diff
@@ -536,3 +536,176 @@ def sigma12_0_cb(self) -> float:
...
+        theta_inn: Optional[float] = None,
+        M_inn: Optional[float] = None,
+    ) -> None:
+        """Initialise the BACCOemu baryon-ratio spline.
+
+        Runs the emulator twice (DMO and baryonic) and stores
+        ``B(z, k)`` as a bivariate spline in ``self._baryon_ratio_interp``.
+
+        Call this **after** ``BACCOemuNonLinearPerturbations.__init__`` so
+        that ``self.emu``, ``self.params_emu``, and ``self.background`` are set.
```
PedroCarrilho: This does not work if we want to apply Bacco baryons to a different nonlinear prescription, e.g. HMCode, right? I.e. this Mixin can only be used with Bacco Nonlinear. You could probably just do what is done elsewhere and use:
`self.emu = emu[nonlinear_model_name][baryonic_model_name]`
and `nonlinear_model_name` can probably be any of the options like `'Arico2023'`

cloelib/cosmology/baccoemu_cosmology.py
```diff
@@ -536,3 +536,176 @@ def sigma12_0_cb(self) -> float:
...
+        }
+        _, Pk_dmo = self.emu.get_nonlinear_pk(
+            cold=False, baryonic_boost=None, k=all_k, **self.params_emu
+        )
+        _, Pk_baryon = self.emu.get_nonlinear_pk(
+            cold=False,
+            baryonic_boost=baryonic_model_name,
+            k=all_k,
+            **{**self.params_emu, **baryonic_params},
+        )
```
PedroCarrilho, line 612: You don't need to get both Pk_dmo and Pk_baryon in Bacco, you can just use directly `emu.get_baryonic_boost`, which is a lot more efficient, since it only needs one emulator, instead of the at least 3 needed to get the full spectrum (linear+nonlinear+baryons)

cloelib/cosmology/baccoemu_cosmology.py
```diff
@@ -536,3 +536,176 @@ def sigma12_0_cb(self) -> float:
...
+        zs = np.atleast_1d(zs)
+        ks = np.atleast_1d(ks)
+        if k_hunit:
+            ks = ks * self.background.h
+        return self._baryon_ratio_interp(zs, ks)
+
+
+class BACCOemuNonLinearBaryonicPerturbations(
+    BACCOemuBaryonBoostMixin, BACCOemuNonLinearPerturbations
+):
```
PedroCarrilho: I understand this is an example of a particular combination that works, but given that these combinations can be automatically created using the function `with_baryon_boost`, I suggest to just use that here, since it is equivalent, right? This is just to avoid confusion of future developers who might think (like I originally did) that they have to build these explicit combinations this way for every case they want to use.

cloelib/cosmology/cosmology.py
```diff
@@ -159,3 +159,131 @@ def matter_power_spectrum_cb(self, zs, ks) -> np.ndarray:
...
+                    zs, ks, k_hunit=kwargs.get("k_hunit", False)
+                )
+            ).squeeze()
+
+        def matter_power_spectrum_cb(self, zs, ks, **kwargs):
+            pk = NonLinearClass.matter_power_spectrum_cb(self, zs, ks, **kwargs)
+            return (
+                pk
+                * self.baryonic_suppression(
+                    zs, ks, k_hunit=kwargs.get("k_hunit", False)
```
PedroCarrilho, line 279: Probably a question for @JegerBroxterman, @MatthieuSchaller or @garico92. Do you think it will matter much if the boost for cb is the same as for the total matter? We could use here  a slightly better approximation, such as the one used to relate the cb to the tot P_NL(k) in Bacco, but this is likely to be such a tiny effect that it is unlikely to matter, right?

cloelib/cosmology/FlamingoBaryonResponseEmulator_cosmology.py
```diff
@@ -0,0 +1,281 @@
...
+            if z <= _Z_MAX_TRAINED:
+                response[i, :], variance[i, :] = (
+                    self.flamingo_emulator.predict_with_variance(
+                        ks_clamped, z, self.fgas, self.Mstar, self.jet
+                    )
+                )
+        return response, variance
+
+
+class CAMBNonLinearFLAMINGOPerturbations(
```
PedroCarrilho: Same point as before, it could be clearer to use here `with_baryon_boost`. Additionally, this seems to be missing the cb P(k), which would automatically get generated when using `with_baryon_boost`

cloelib/cosmology/HMcode2020Emu_cosmology.py
```diff
@@ -328,3 +328,125 @@ def _set_neutrino_masses(background: Background) -> float:
...
+        if k_hunit:
+            ks = ks * self.background.h
+        result = np.ones((len(zs), len(ks)))
+        in_range = ks >= self._baryon_k_nl_min
+        if np.any(in_range):
+            result[:, in_range] = self._baryon_ratio_interp(zs, ks[in_range])
+        return result
+
+
+class HMemuNonLinearBaryonicPerturbations(
```
PedroCarrilho: Same point here as the previous one, with the added issue that this combination already exists, i.e. it is possible to get HMemu NL + Baryons with the usual method, which directly uses the emulator for P(k) with baryons and is more efficient. As such, this combination may be removed here. If an example would be useful to have, you can select a different NL model, e.g. from jax_cosmology or Bacco.

cloelib/cosmology/baccoemu_cosmology.py
```diff
@@ -536,3 +536,176 @@ def sigma12_0_cb(self) -> float:
...
+    .. note::
+        ``__init__`` must be called **after** the base NonLinear ``__init__``
+        because it reads ``self.emu``, ``self.params_emu``, and
+        ``self.background`` which are set there.
+    """
+
+    def __init__(
+        self,
+        baryonic_model_name: str = "Burger2025",
+        M_c: Optional[float] = None,
```
PedroCarrilho: These parameters are optional, but there doesn't seem to be a default option for what happens when they are not passed (I guess an error is just raised later). Should these be made non-optional? Or raise a more specific error?

cloelib/cosmology/FlamingoBaryonResponseEmulator_cosmology.py
```diff
@@ -0,0 +1,281 @@
...
+    zero-boilerplate class creation.
+    """
+
+    # ------------------------------------------------------------------
+    # Initialiser
+    # ------------------------------------------------------------------
+
+    def __init__(
+        self,
+        fgas_sigma: float = 0.0,
```
PedroCarrilho, line 68: Similar comment on default parameter values. In this case all of them are valid, which means no error will appear. Is this the behaviour we want?

PedroCarrilho (commented):
This is looking really great! I left a few comments which we probably have to resolve, but I don't think they are very big issues at all.

One last thing that is valid for all models: as far as I can tell (but I am a bit tired), there are no extrapolations in k and z or they are resolved either via each package independently or via the default extrapolation of the interpolators used here (typically constant). This may be fine, but we could also use the `extend_spectra` function from `cloelib.auxiliary.extrapolator` to control this explicitly. I can help doing that.

cloelib/cosmology/baccoemu_cosmology.py
gcanasherrera: Not sure what the best approach is, but we will need to homogenise pretty much everywhere. What's the natural behaviour of the emulator itself @PedroCarrilho?

cloelib/cosmology/baccoemu_cosmology.py
gcanasherrera: Changed.

cloelib/cosmology/baccoemu_cosmology.py
PedroCarrilho: The emulator itself always needs these parameters, so if you pass None it will throw an error, e.g. `KeyError: ('baryon emulator: coordinates need the following parameters: ', ['M_c'])`, i.e. there are no emulator defaults. 

So, they are not really optional, unless we select some defaults ourselves. My opinion is that they should not be optional, even with a default, because I don't think we have an ideal option for what the default should be and I don't think we want to bias users in using the particular default set of parameters we choose as "the Bacco prediction". I think having to input some parameters will give users the sense that all these predictions are parameter-dependent.

That being said, two of these parameters are optional (`theta_out` and `M_inn`), since they are only used in the Arico 2021 model, but not in the Burger 2025. So, IMO, those should remain optional and set to None, to avoid any silent failures should the user select the Arico 2021 model.

I think the same should be true in all other baryon models, i.e. without any optional parameters.

cloelib/cosmology/baccoemu_cosmology.py
```diff
@@ -610,19 +602,13 @@ def __init__(
...
         }
-        _, Pk_dmo = self.emu.get_nonlinear_pk(
-            cold=False, baryonic_boost=None, k=all_k, **self.params_emu
-        )
-        _, Pk_baryon = self.emu.get_nonlinear_pk(
-            cold=False,
-            baryonic_boost=baryonic_model_name,
-            k=all_k,
-            **{**self.params_emu, **baryonic_params},
+        k_baryon = baryon_emu.emulator["baryonic"]["k"]
```
PedroCarrilho: This should be `"baryon"` instead of `"baryonic"`. Also, if doing it this way, you don't actually need to get `k_baryon` from here. You can just not pass it below and then get it from the output of `get_baryonic_boost` as:
```
k_baryon, boost = baryon_emu.get_baryonic_boost(
           **{**self.params_emu, **baryonic_params}
)
```

Beyond this, note that now choosing to produce the output only for k_baryon instead of `all_k` means that you are using the default extrapolation at low k from `RectBivariateSpline`, which defaults to constant. This is actually what we want, since the boost is 1 at large scales, but could sometimes be unpredictable, so we may want to change all extrapolations.

cloelib/cosmology/cosmology.py
gcanasherrera: @JegerBroxterman @MatthieuSchaller @garico92 could anyone give us a hand here? :)

cloelib/cosmology/HMcode2020Emu_cosmology.py
gcanasherrera: I removed this class

cloelib/cosmology/baccoemu_cosmology.py
gcanasherrera:
```
        M_c: float = 0.0,
        eta: float = 0.0,
        beta: float = 0.0,
        M1_z0_cen: float = 0.0,
        theta_out: Optional[float] = None,
        M_inn: Optional[float] = None,
        theta_inn: float = 0.0,
```

?

cloelib/cosmology/baccoemu_cosmology.py
gcanasherrera: I'm quite confused about the extrapolations at this stage. You have a stronger experience on this. What if we open a follow up task tackling as well the extrapolator.py ?

gcanasherrera: I updated this part of the code though.

gcanasherrera:
@PedroCarrilho I've changed the initialization of some parameters following your advice above (but, to which values?? for the moment all to zero).

We have pending your question about _cb and extrapolations. But I think the best for extrapolations is to do it in a follow up PR.

gcanasherrera:
(apologies for closing, wrong click :D !! )

cloelib/cosmology/cosmology.py
MatthieuSchaller: What do you mean by `boost for cb`?

In our case, the boost is (baryons + CDM + neutrinos) / (CDM + neutrinos).

garico92: Hi all! @MatthieuSchaller I think that @PedroCarrilho is referring to the fact that other models, like BACCOemu, predict the “cold" baryonic boost, that is (baryons + CDM) / (CDM + neutrinos). He was suggesting that maybe one can improve slightly on this by adding neutrinos at a linear level (as done in BACCOemu nonlinear boost). 
I think that could be quite easily done, but in practical terms the correction would be negligible in all cases (I just made a small example with Mnu=0.4 eV and f_nu=  0.04, differences were of the order of 10^-5). 
We have also checked with FLAMINGO sims that differences between total and cold baryonic boosts are negligible, so I’d say this is not an issue.

MatthieuSchaller: Right, thanks for the clarification, Giovanni.

Yes, I agree that the difference will be negligible in any practical scenario. Might be worth mentioning this somewhere in the documentation as a caveat though. 

We may want to coordinate a bit more the interface and definitions of future emulators we produce.

cloelib/cosmology/baccoemu_cosmology.py
garico92: As Pedro said, the BACCOemu philosophy was to not provide default values for the parameters, so that the user is “responsabilized” of their choices. I don’t know if the rest of the library provides default values, I think one risk of having them for baryons is that, if the common user has an idea of what is h0 or ns, and broadly knows the value their values, that is probably not the case for parameters such as M_inn or theta_out, thus they would take the defaults acritically.

cloelib/cosmology/baccoemu_cosmology.py
garico92: I think that extrapolating or fixing to 1 on large scales make little difference, but on small scales the power 
law extrapolation would work much better than other extrapolations

cloelib/cosmology/cosmology.py
PedroCarrilho: Thanks everyone for your input. @gcanasherrera, I think the easiest thing to do is to just leave it as it is and add a comment saying that we use the same boost for total and cb, as this is a very good approximation.

cloelib/cosmology/baccoemu_cosmology.py
PedroCarrilho: Hi @gcanasherrera, I think I confused my wording a little bit. I didn't mean you should drop the `Optional[float]` (I believe that only takes care of the type, right?). I meant you should not provide a default value, i.e. make passing the variable non-optional, so that it would be:

```
        M_c: Optional[float],
        eta: Optional[float],
        beta: Optional[float],
        M1_z0_cen: Optional[float],
        theta_inn: Optional[float],
        theta_out: Optional[float] = None,
        M_inn: Optional[float] = None,
```
whether we need the `Optional` or not, I leave it to you.

Overall, this is not a big difference to what you had before, since the code will later throw an error if you pass None (using 0 as a default is not great, since, for parameters eta, beta and theta_inn that is valid, but may not be what people want). The only small change in what I am suggesting is that it would throw an error at class initialisation immediately, since some parameters were not passed. In any case, if you prefer, for any reason, to have a default value for each of these, I would pass None, as you initially had.

cloelib/cosmology/baccoemu_cosmology.py
PedroCarrilho: Indeed as Giovanni says, the problem is the smaller scales, since now the extrapolation defaults to constant at high k, which is a bit far from the expected behaviour. If you want, I can do this in a different task, since it requires a few extra details. For example, to do a good extrapolation, we need to pass to these classes the values of k for which we want to extrapolate or decide on a pre-defined range, and we are not doing that now (also more generally in other parts of the code, including in EE2, where I forgot...). Similarly, we also need to pass here the requested redshifts to extrapolate.

cloelib/cosmology/cosmology.py
gcanasherrera: I will add a comment and update the docs, cheers @PedroCarrilho and everyone!

MatthieuSchaller:
Hi guys,

I am also interested in providing the analytic correction we had in https://ui.adsabs.harvard.edu/abs/2025MNRAS.540.2322S/abstract as an optional module. 
I was waiting for this PR to be merged so that I know the format to add another module is finalised, but I could do it here  too. 
Are there still big hurdles to complete this PR? And if so, anything I can help with?

MatthieuSchaller:
Hi all,
Is there anything we can do here to help further with the integration of the baryon corrections?

gcanasherrera:
> Hi all, Is there anything we can do here to help further with the integration of the baryon corrections?

Hi @MatthieuSchaller, the reason why I haven't merged it is because we are working on an internal restructuring to enable alternative intrinsic alignment models + non-linear galaxy bias + beyond Limber integration. Before merging all of this, I wanted to make sure that the new structure that we have for Baryonic boosts still holds for this new architecture, which is currently under review at #373.

I do not foresee issues, so I will start pulling those changes here to have an idea how it might look like eventually.

PR #337, opened 2025-11-10
gh-145: Add BNT transformation

### LINKED ISSUE #145: Add BNT transformation
## 📝 Issue Summary
I want to add the BNT. However since the BNT transformation assumes a fiducial cosmology it make sense to compute the BNT with the same cosmology as the covariance matrix for consistency

- [ ] **Goal 1**: Be able to read the cosmology that was used to compute the covariance
- [ ] **Goal 2**: compute the corresponding BNT transform and transform the covariance, data vector and theory vector consistently.

### 👥 Assignees
- @TBA

pltaylor16:
### ✅ Summary

 The purpose is to add the computation of the BNT matrix into cloelib. These will be used in cloelike to build BNT likelihoods.

### 🔄 Changes

Added code to compute the BNT matrix. This will be used by cloelike. 

### 🛠 How to Test

 I have added code so that pytest will automatically test that the BNT matrix is computed correctly.

### 📝 Documentation

- [x] This PR updates documentation 

### 🏗 Related Issues

Resolves #145

### 📌 Additional Notes

@all-contributors please add @pltaylor16 for code, doc

---

### DIFF
--- a/cloelib/auxiliary/bnt.py
+++ b/cloelib/auxiliary/bnt.py
@@ -23,7 +23,7 @@ class BNTMatrixCalculator:
         Parameters
         ----------
         dndz_list : List[np.ndarray or jnp.ndarray]
-            A n-dimensional array representing the number density distribution of galaxies as a function of redshift.
+            A list of 1d-arrays representing the number density distribution of galaxies as a function of redshift.
             Each list member is expected to be normalised.
         z : np.ndarray or jnp.ndarray
             A 1-dimensional array representing the redshift values corresponding to the `dndz` array.
--- a/cloelib/auxiliary/bnt.py
+++ b/cloelib/auxiliary/bnt.py
@@ -46,10 +46,12 @@ class BNTMatrixCalculator:
             raise ValueError(
                 "One of the z array elements is equal to zero, breaking the BNT computation."
             )
-        if len(self.z.shape) != len(self.dndz_list[0].shape):
-            raise ValueError(
-                "The shape of the z array does not match the shape of the dndz array."
-            )
+        for nz in self.dndz_list:
+            if nz.shape != self.z.shape:
+                raise ValueError(
+                    f"Each dndz array must have shape {self.z.shape}, "
+                    f"but found {nz.shape}."
+                )
 
         self.chi = self.background.comoving_distance(self.z)
 
@@ -59,8 +61,16 @@ class BNTMatrixCalculator:
         B_list = []
         for i in range(self.nbins):
             nz = self.dndz_list[i]
-            A_list += [np.trapz(nz, self.z)]
-            B_list += [np.trapz(nz / self.chi, self.z)]
+            A_list.append(
+                np.trapezoid(nz, self.z)
+                if hasattr(np, "trapezoid")
+                else np.trapz(nz, self.z)
+            )
+            B_list.append(
+                np.trapezoid(nz / self.chi, self.z)
+                if hasattr(np, "trapezoid")
+                else np.trapz(nz / self.chi, self.z)
+            )
 
         BNT_matrix = np.eye(self.nbins)
         BNT_matrix[1, 0] = -1.0
@@ -70,7 +80,7 @@ class BNTMatrixCalculator:
                 [[A_list[i - 1], A_list[i - 2]], [B_list[i - 1], B_list[i - 2]]]
             )
             A = -1.0 * np.array([A_list[i], B_list[i]])
-            soln = np.dot(np.linalg.inv(mat), A)
+            soln = np.linalg.solve(mat, A)
             BNT_matrix[i, i - 1] = soln[0]
             BNT_matrix[i, i - 2] = soln[1]
 
--- a/cloelib/auxiliary/bnt.py
+++ b/cloelib/auxiliary/bnt.py
@@ -5,7 +5,9 @@ Module with one class to compute the BNT matrix.
 import numpy as np
 import jax.numpy as jnp
 from cloelib.cosmology.cosmology import Background
-from typing import List, Union
+from typing import List, TypeVar, Union
+
+T = TypeVar("T", bound=Union[jnp.ndarray, np.ndarray])
 
 
 class BNTMatrixCalculator:
@@ -13,8 +15,8 @@ class BNTMatrixCalculator:
 
     def __init__(
         self,
-        dndz_list: List[Union[np.ndarray, jnp.ndarray]],
-        z: Union[np.ndarray, jnp.ndarray],
+        dndz_list: List[T],
+        z: T,
         background: Background,
     ):
         r"""
--- a/cloelib/auxiliary/bnt.py
+++ b/cloelib/auxiliary/bnt.py
@@ -92,7 +92,13 @@ class BNTMatrixCalculator:
                 [[A_list[i - 1], A_list[i - 2]], [B_list[i - 1], B_list[i - 2]]]
             )
             A = -1.0 * np.array([A_list[i], B_list[i]])
-            soln = np.linalg.solve(mat, A)
+            try:
+                soln = np.linalg.solve(mat, A)
+            except np.linalg.LinAlgError as exc:
+                raise ValueError(
+                    "BNT matrix construction failed: non-invertible 2x2 "
+                    f"system encountered at tomographic bin index {i}."
+                ) from exc
             BNT_matrix[i, i - 1] = soln[0]
             BNT_matrix[i, i - 2] = soln[1]
 
--- a/cloelib/auxiliary/bnt.py
+++ b/cloelib/auxiliary/bnt.py
@@ -71,31 +71,21 @@ class BNTMatrixCalculator:
             The BNT transformation matrix with shape ``(nbins, nbins)``, always
             lower triangular with ones on the diagonal.
         """
-        A_list = []
-        B_list = []
-        for i in range(self.nbins):
-            nz = self.dndz_list[i]
-            A_list.append(
-                np.trapezoid(nz, self.z)
-                if hasattr(np, "trapezoid")
-                else np.trapz(nz, self.z)
-            )
-            B_list.append(
-                np.trapezoid(nz / self.chi, self.z)
-                if hasattr(np, "trapezoid")
-                else np.trapz(nz / self.chi, self.z)
-            )
+
+        trapz = np.trapezoid if hasattr(np, "trapezoid") else np.trapz
+
+        nz_arr = np.stack(self.dndz_list, axis=0)
+        A = trapz(nz_arr, self.z, axis=1)
+        B = trapz(nz_arr / self.chi[None, :], self.z, axis=1)
 
         BNT_matrix = np.eye(self.nbins)
         BNT_matrix[1, 0] = -1.0
 
         for i in range(2, self.nbins):
-            mat = np.array(
-                [[A_list[i - 1], A_list[i - 2]], [B_list[i - 1], B_list[i - 2]]]
-            )
-            A = -1.0 * np.array([A_list[i], B_list[i]])
+            mat = np.array([[A[i - 1], A[i - 2]], [B[i - 1], B[i - 2]]])
+            rhs = -np.array([A[i], B[i]])
             try:
-                soln = np.linalg.solve(mat, A)
+                soln = np.linalg.solve(mat, rhs)
             except np.linalg.LinAlgError as exc:
                 raise ValueError(
                     "BNT matrix construction failed: non-invertible 2x2 "
--- a/cloelib/auxiliary/bnt.py
+++ b/cloelib/auxiliary/bnt.py
@@ -1,0 +1,89 @@
+# file: bnt.py
+import numpy as np
+from cloelib.cosmology.cosmology import Background
+from typing import List
+
+class BNT:
+    """Class to compute the BNT matrix."""
+
+    def __init__(
+        self,
+        dndz_list : List[np.ndarray],
+        z : np.ndarray,
+        fid_parameters: dict,
+        background : Background,
+    ):
+        r"""
+        Initialize the class instance.
+
+        Parameters
+        ----------
+        dndz_list : List[np.ndarray]
+            A n-dimensional array representing the number density distribution of galaxies as a function of redshift.
+            Each list member is expected to be normalised.
+        z : np.ndarray
+            A 1-dimensional array representing the redshift values corresponding to the `dndz` array.
+        fid_parameters : dict
+            A dictionary of the fiducial cosmology parameters.
+        background : Background
+            Cosmological background instance.
+        """
+
+        self.dndz_list = dndz_list
+        self.z = z
+        self.fid_parameters = fid_parameters
+        self.background = background
+        self.nbins = len(self.dndz_list)
+
+        if len(self.dndz_list) < 3:
+            raise ValueError(
+                "BNT requires at least 3 tomographic bins to compute the matrix."
+            )
+        if np.any(self.z == 0.0):
+            raise ValueError(
+                "One of the z array elements is equal to zero, breaking the BNT computation."
+            )
+        if len(self.z.shape) != len(self.dndz_list[0].shape):
+            raise ValueError(
+                "The shape of the z array does not match the shape of the dndz array."
+            )
+        self.chi  = self.background.comoving_distance(z)
+        
+
+    def get_matrix(self):
+        """Compute the BNT matrix."""
+        A_list = []
+        B_list = []
+        for i in range(self.nbins):
+            nz = self.dndz_list[i]
+            A_list += [np.trapz(nz, self.z)]
+            B_list += [np.trapz(nz / self.chi, self.z)]
+
+
+        BNT_matrix = np.eye(self.nbins)
+        BNT_matrix[1,0] = -1.
+
+        for i in range(2,self.nbins):
+            mat = np.array([ [A_list[i-1], A_list[i-2]], [B_list[i-1], B_list[i-2]] ])
+            A = -1. * np.array( [A_list[i], B_list[i]] )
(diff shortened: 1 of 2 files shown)


gcanasherrera:
@all-contributors please add @pltaylor16 for code, doc (@pltaylor16 note that this needs to be added in a comment :) )

gcanasherrera:
Also @pltaylor16, make sure you have a catchy pic in your profile!

pltaylor16:
@marcobonici @davidesciotti @santiagocasas I have just pushed another commit on the branch to ensure the BNT matrix computation is compatible with both jnp and np datatypes.

marcobonici:
Thanks @pltaylor16 ! We are now under going review:)

cloelib/auxiliary/bnt.py
```diff
@@ -0,0 +1,141 @@
...
+            A 1-dimensional array representing the redshift values corresponding to the `dndz` array.
+        fid_parameters : dict
+            A dictionary of the fiducial cosmology parameters.
+        background : Background
+            Cosmological background instance.
+        """
+
+        # --- backend selection & consistency check ---
+        xp = _pick_xp(z)
+        z_is_jax = _is_jax_array(z)
```
marcobonici: @pltaylor16 , can you explain to me the rationale behind this choice? We are not jitting this part of the computation, so we can deal with numpy arrays.

cloelib/auxiliary/bnt.py
```diff
@@ -0,0 +1,141 @@
...
+            raise ValueError(
+                "One of the z array elements is equal to zero, breaking the BNT computation."
+            )
+        if len(self.z.shape) != len(self.dndz_list[0].shape):
+            raise ValueError(
+                "The shape of the z array does not match the shape of the dndz array."
+            )
+
+        # background typically returns NumPy; convert to selected backend
+        self.chi = self.background.comoving_distance(z)
```
marcobonici: Same as above, why converting to numpy?

cloelib/auxiliary/bnt.py
```diff
@@ -0,0 +1,141 @@
...
+    """Backend-agnostic trapezoidal integration."""
+    if xp is np:
+        return np.trapz(y, x, axis=axis)
+    else:
+        # JAX version
+        dx = x[1:] - x[:-1]
+        # reshape dx for broadcasting
+        sl = [None] * y.ndim
+        sl[axis] = slice(None)
+        dx = dx[tuple(sl)]
```
marcobonici: Also, why using trapz for this integral? We should be able to use the simpson weights, that are coded in the auxiliary/math_utils. If this specific function is required, let us code it there (so that it can be used by other people eventually :) ).

marcobonici (changes requested):
Hi @pltaylor16 , I went through your PR, it looks good!
Before proceeding, I have a couple of points raised:
- **Forcing the consistency of numpys**. Why are you requiring the user to be using always jax or plain numpy? SInce this part of the computation is done only once, we can just avoid jitting it, it will be a bit slower but it will be fine.
- **Math utils**. You wrote a funciton for trapz integration. We already have a simpson based on in math_utils. So, I would advise to either use that or to implement the numpy trapz one there (if there is a reason to prefer trapz to simps).

pltaylor16:
@marcobonici I have just pushed a new commit to the branch. The new behavior is that the class will accept numpy and jnp arrays and always return a numpy array with computations performed using numpy. The earlier clumsier implementation was because I didn't realise that jax will always lift constants from the graph when running JIT, so feeding always feeding in a numpy array to the likelihood computation will never be a problem.

cloelib/auxiliary/bnt.py
pltaylor16: see response below.

pltaylor16: see response below

cloelib/auxiliary/bnt.py
pltaylor16: see response below.

cloelib/auxiliary/bnt.py
pltaylor16: see response below

cloelib/auxiliary/bnt.py
```diff
@@ -0,0 +1,81 @@
...
+Module with one class to compute the BNT matrix.
+"""
+
+import numpy as np
+import jax.numpy as jnp
+from cloelib.cosmology.cosmology import Background
+from typing import List, Union
+
+
+class BNT:
```
marcobonici: This is minor, but do you think you could have a longer name for this class?

cloelib/auxiliary/bnt.py
```diff
@@ -0,0 +1,81 @@
...
+                "One of the z array elements is equal to zero, breaking the BNT computation."
+            )
+        if len(self.z.shape) != len(self.dndz_list[0].shape):
+            raise ValueError(
+                "The shape of the z array does not match the shape of the dndz array."
+            )
+
+        self.chi = self.background.comoving_distance(self.z)
+
+    def get_matrix(self) -> np.ndarray:
```
marcobonici: I think the name `get_matrix` is a bit short. THis can be extended to `get_BNT_matrix` maybe...?

marcobonici (changes requested):
Hi @pltaylor16 , I made another couple of small comment.

cloelib/auxiliary/bnt.py
pltaylor16: see below

cloelib/auxiliary/bnt.py
pltaylor16: see below

pltaylor16:
@marcobonici I have made the changes you suggested.
@davidesciotti @santiagocasas Please let me know what you think.

cloelib/auxiliary/bnt.py
```diff
@@ -0,0 +1,81 @@
...
+        fid_parameters: dict,
+        background: Background,
+    ):
+        r"""
+        Initialize the class instance.
+
+        Parameters
+        ----------
+        dndz_list : List[np.ndarray or jnp.ndarray]
+            A n-dimensional array representing the number density distribution of galaxies as a function of redshift.
```
marcobonici: Shouldn't this just be a 2D array? (Sorry, I should have catched that before).

cloelib/auxiliary/bnt.py
```diff
@@ -0,0 +1,81 @@
...
+            A dictionary of the fiducial cosmology parameters.
+        background : Background
+            Cosmological background instance.
+        """
+
+        # Convert everything to NumPy for internal computations
+        self.z = np.asarray(z)
+        self.dndz_list = [np.asarray(nz) for nz in dndz_list]
+
+        self.fid_parameters = fid_parameters
```
marcobonici: Why is here a `fid_parameters` being defined, since it doesn't look to me it is never used?

cloelib/auxiliary/bnt.py
marcobonici: Ok, you are using a list of 1D arrays. Please, update the docstring accordingly.

marcobonici (changes requested):
Hi @pltaylor16 , on my side we should be there. I have asked for a small thing on the docstring side and for a clarification why we have a fiducial_cosmology if that is never used. Once these are fixed, greenlight from my side.
Maybe also @davidesciotti can give a look, as he worked with BNT before...?
Tagging @gcanasherrera as well.

cloelib/auxiliary/bnt.py
pltaylor16: Resolved.

cloelib/auxiliary/bnt.py
pltaylor16: resolved.

pltaylor16:
@marcobonici All outstanding issues have been resolved. Thank you for reviewing.

davidesciotti:
@marcobonici @pltaylor16 Very sorry for the late reply, I'm taking a look at the PR now

cloelib/auxiliary/bnt.py
```diff
@@ -0,0 +1,77 @@
...
+
+        if len(self.dndz_list) < 3:
+            raise ValueError(
+                "BNTMatrixCalculator requires at least 3 tomographic bins to compute the matrix."
+            )
+        if np.any(self.z == 0.0):
+            raise ValueError(
+                "One of the z array elements is equal to zero, breaking the BNT computation."
+            )
+        if len(self.z.shape) != len(self.dndz_list[0].shape):
```
davidesciotti: About `if len(self.z.shape) != len(self.dndz_list[0].shape)`:
I'm a bit confused by this check, `len(self.z.shape)` is actually the number of dimensions (1) of the array, so this is not checking that the number of elements in the z array and `dndz_list[0] `is the same... But maybe I'm misunderstanding the purpose

cloelib/auxiliary/bnt.py
```diff
@@ -0,0 +1,77 @@
...
+        self.chi = self.background.comoving_distance(self.z)
+
+    def get_bnt_matrix(self) -> np.ndarray:
+        """Compute the BNT matrix."""
+        A_list = []
+        B_list = []
+        for i in range(self.nbins):
+            nz = self.dndz_list[i]
+            A_list += [np.trapz(nz, self.z)]
+            B_list += [np.trapz(nz / self.chi, self.z)]
```
davidesciotti: `np.trapz` has been deprecated in favour of `np.trapezoid` since numpy v2.0. I'm not sure whether we require an older version but this could create compatibility issues in the future

cloelib/auxiliary/bnt.py
```diff
@@ -0,0 +1,77 @@
...
+
+        BNT_matrix = np.eye(self.nbins)
+        BNT_matrix[1, 0] = -1.0
+
+        for i in range(2, self.nbins):
+            mat = np.array(
+                [[A_list[i - 1], A_list[i - 2]], [B_list[i - 1], B_list[i - 2]]]
+            )
+            A = -1.0 * np.array([A_list[i], B_list[i]])
+            soln = np.dot(np.linalg.inv(mat), A)
```
davidesciotti: `soln = np.linalg.solve(mat, A)` should be more numerically stable than inverting and then using `np.dot`.
Also, maybe it would be good to add a check, something like

```
try:
    soln = np.linalg.solve(mat, A)
except np.linalg.LinAlgError as e:
    raise ValueError(
        f'Failed to solve linear system for bin {i}: '
        'matrix is singular or ill-conditioned. '
    ) from e

File: tests/test_bnt.py [hunk 11]
```diff
@@ -0,0 +1,160 @@
...
+                [1.0, 0.0, 0.0],
+                [-1.0, 1.0, 0.0],
+                [0.2969388, -1.2969388, 1.0],
+            ]
+        )
+        assert_allclose(BNT_matrix, expected, rtol=0, atol=1e-4)
+
+    def test_bnt_matrix_correct_to_4dp_jax(self):
+        """Matrix matches expected values to 4 decimal places when inputs are JAX arrays."""
+        import numpy as np
```
davidesciotti: Redundant imports?

cloelib/auxiliary/bnt.py
```diff
@@ -0,0 +1,77 @@
...
+            raise ValueError(
+                "One of the z array elements is equal to zero, breaking the BNT computation."
+            )
+        if len(self.z.shape) != len(self.dndz_list[0].shape):
+            raise ValueError(
+                "The shape of the z array does not match the shape of the dndz array."
+            )
+
+        self.chi = self.background.comoving_distance(self.z)
+
```
davidesciotti, line 59: Missing documentation on what the function returns (the type hint might be enough, just pointing out if you want to be more specific) 
[I'm referring to the `get_bnt_matrix` function, I clearly have some issue in selecting the appropriate lines of code for the review comments :) ]

cloelib/auxiliary/bnt.py
pltaylor16: Thanks, good catch. I've updated the code to check each of the dndz in the list match the length of the z array.

cloelib/auxiliary/bnt.py
pltaylor16: updated code so that it will run with np.trapz or np.trapezoid.

cloelib/auxiliary/bnt.py
pltaylor16: changed to solve as suggested.

I haven't added the error statement. If it is throwing an error message at np.linalg.solve then the user is going to have to dig really deep, so I don't adding this one extra error message will help much. I have used BNT a lot, so I can't imagine a situation during normal use where you will get an error here in any case.

tests/test_bnt.py
pltaylor16: fixed

cloelib/auxiliary/bnt.py
pltaylor16: I have left this as is. I think this is already clear, and we also don't want too many docstrings floating around.

marcobonici: I support @davidesciotti 's pov. The docstring for the `get_bnt_matrix` can be extended.

cloelib/auxiliary/bnt.py
marcobonici: The point about this kind of test is to prevent issues to go undetected.
I think that a good thing to do would be to include a unit test, showing that the code is throwing an error if the system is singular. IT should be fairly easy: create a (physically not meanginful) ill-defined matrix and feed it to the code. The unit test should simply check that the code is throwing an error (which should be the case, since we are relying on numpy).

pltaylor16: @marcobonici The code was updated to use np.linalg.solve to compute the elements in the last commit rather than np.linalg.inv. np.linalg.solve will already throw a singular matrix error if the matrix is singular, so I think these extra lines would be superfluous.

cloelib/auxiliary/bnt.py
pltaylor16: docstrings updated in latest commit.

cloelib/auxiliary/bnt.py
marcobonici: Hi @pltaylor16 , I see your point, but the idea is that if in the future there is any issue, this will raise a flag.
I agree that with the actual implementation, based on `numpy`, this is not an issue but this, IMHO, is the whole point of the unit tests. For the comoving distance, for instance, I have a unit test checking that for $z=0$, the comoving distance is 0. This is _obvious_ and the unit test from the point of view might look superfluous, but I think it is better to have such a test (that is nevertheless very short and quick to write) rather than not having it.
So: could you please add a few lines of code creating a (non physical) singular BNT matrix and checking that an error is raised, please?

pltaylor16: @marcobonici @davidesciotti Okay. I have added a line which will throw and error if the matrix is not invertible and I have added a unit test.

pltaylor16:
@marcobonici @davidesciotti Added some code to throw an error if the matrix is not invertible and a unit test to make sure this works. I think there are no further outstanding issues. @santiagocasas I believe that we are just waiting for your review.

cloelib/auxiliary/bnt.py
```diff
@@ -0,0 +1,105 @@
...
+from cloelib.cosmology.cosmology import Background
+from typing import List, Union
+
+
+class BNTMatrixCalculator:
+    """Class to compute the BNT matrix."""
+
+    def __init__(
+        self,
+        dndz_list: List[Union[np.ndarray, jnp.ndarray]],
```
gcanasherrera: Check out cosmology.py, in particular the Background protocol, to define a new variable T as union.

cloelib/auxiliary/bnt.py
```diff
@@ -0,0 +1,105 @@
...
+        weak-lensing kernels, designed to construct linear combinations of shear kernels
+        that are localised in comoving distance.
+
+        Returns
+        -------
+        np.ndarray
+            The BNT transformation matrix with shape ``(nbins, nbins)``, always
+            lower triangular with ones on the diagonal.
+        """
+        A_list = []
```
gcanasherrera: Why these for-loops?

```
# Just to check, because in principle all of these are np/jpn arrays
dndz = np.asarray(self.dndz_list)
z = np.asarray(self.z)
chi = np.asarray(self.chi)

# Note that you're assuming numpy > 2., we are restricted by tensorflow in HMCode2020Emu tu numpy < 2
trapz = np.trapezoid if hasattr(np, "trapezoid") else np.trapz

A_list = trapz(dndz, z, axis=1)
B_list = trapz(dndz / chi[None, :], z, axis=1)
```

gcanasherrera (changes requested):
Sorry for the delay, a couple of things only

cloelib/auxiliary/bnt.py
pltaylor16: I have just updated code to define a new variable T as a Union.

cloelib/auxiliary/bnt.py
pltaylor16: The elements of the matrix are computed row by row in my code. The BNT matrix is only computed once, so I have not vectorised the computation.

gcanasherrera: I should be straightforward so let's try to be efficient just in case, thanks! And happy new year!

pltaylor16: @gcanasherrera I have made the code as efficient as possible in the last push. I have vectorised the first for loop, but since computing the BNT elements is fundamentally sequential, I see no way to remove the 2nd for loop.

pltaylor16:
@santiagocasas @davidesciotti @marcobonici @gcanasherrera  I think the code should now be ready to merge. Please let me know if you think I need to make any additional changes.

marcobonici (approved):
Hi @pltaylor16 , I am happy with the status. I am merging!

PR #352, opened 2025-11-20
gh-292: Incorporate mochi_class

### LINKED ISSUE #292: Implement mochi_class
## 📝 Issue Summary

Implementing <a href="https://github.com/mcataneo/mochi_class_public">mochi_class</a>  into cloelib

- [x] **Goal 1**: Create a new protocol file for mochi_class in /cosmology for the implementation of background and perturbations using mochi_class
- [x] **Goal 2**: Define the input parameters of mochi_class in the initialization section, including the name of the chosen model and a dictionary to pass the evolution array of the stable basis functions
- [x] **Goal 3**: Adapt the names for cosmological quantitites for CLASS v3.2.2
- [ ] **Goal 4**: Run chains to test the implementation 

### ✅ Expected Outcome

Being able to pass mochi_class input parameters to CLOE and use the stable basis to compute the background and perturbation cosmology in CLOE

### 🐞 Current Behavior 

There's currently no way to load the stable basis functions in cloelib.

### 📌 Additional Notes

The mochi_class version of classy has to be the first in the path when using mochi_class for the background and perturbations in`cloelib`.

### 👥 Assignees

- @cosmicLinux 
- @LisaGoh 
- @mcataneo

LisaGoh:
### ✅ Summary

Incorporate mochi_class into cloelib, and add the stable basis parametrisation following Lombriser et al 2019

### 🔄 Changes

- Added the mochiclass Background and Perturbation classes in the mochi_class_cosmology.py script
- Included a mochi_stable_param.py script to derive the related mochi quantities given a set of stable parameters
- Updated the photo.py to be compatible with the general Class Perturbation class

### 📝 Documentation

### 🏗 Related Issues

Resolves #292 

### 📌 Additional Notes

---


cloelib/observables/photo.py
```diff
@@ -81,7 +81,7 @@ def get_window_IA(self, z):
         -------
         window_IA: np.ndarray
         """
-        Omega_m0 = self.background.Omega_m(0.0)
+        Omega_m0 = self.background.Omega_m([0.0])
```
gcanasherrera: why have you changed this?

cloelib/cosmology/mochi_class_cosmology.py
```diff
@@ -0,0 +1,718 @@
...
+from cloelib.cosmology.cosmology import Background
+from cloelib.auxiliary.units import SPEED_OF_LIGHT
+import sys
+
+# General imports
+import numpy as np
+import copy
+from typing import Optional, Union, Sequence
+
+np.set_printoptions(threshold=sys.maxsize)
```
gcanasherrera, line 14: why this?

cloelib/cosmology/mochi_class_cosmology.py
```diff
@@ -0,0 +1,718 @@
...
+
+# Cosmology imports (make sure it's the version of mochi_class that's being imported not the standard CLASS!)
+try:
+    from classy import Class  # type: ignore
+except ImportError as e:
+    raise ImportError("classy could not be imported.") from e
```
gcanasherrera: Would be worth printing the path to the classy imported, just because the name is the same. Alternatively @mcataneo can we actually update the name of mochi_class python wrapper please?

cloelib/cosmology/mochi_stable_param.py
```diff
@@ -0,0 +1,226 @@
+# General imports
```
gcanasherrera: I know this file is important for the project, but technically it does not live within the cloelib package. I'm thinking about options, let's chat in person about this.

gcanasherrera (changes requested):
Remove the mochi_stable_params.py file. Also, check some other comments for the moment.

cloelib/observables/photo.py
LisaGoh: I changed this in order to be compatible with how `class_cosmology.py` was coded (on which `mochi_class_cosmology.py` is based). The `Omega_m(z)` method in line 251 of `class_cosmology.py` goes as

`return np.array([self.results.Om_m(z) for z in zs])`

hence would return an input type error if it is called as `Omega_m0 = self.background.Omega_m(0.0)` in `photo.py` (the input being a float and not an array).

cloelib/cosmology/mochi_stable_param.py
LisaGoh: Yes indeed I was on the fence about this. I put it here as a temporary solution so that others can run the notebooks as well, but more than happy to change the structure!

cloelib/cosmology/mochi_class_cosmology.py
LisaGoh: Currently the MG parameters of `mochiclass_stable_basis_dict` are being read in as ` "lna_smg": np.array2string(lna_smg, separator=",", precision=16).replace("\n", "").strip("[]")` and so on. Without this `np.set_printoptions` settings updated, the `lna_smg`, `Delta_M2`, `D_kin` and `cs2` arrays are not being printed out in full and formatted properly, and mochi_class throws out an input error. 

This was what I learned from debugging and printing out the output after each line, but it's probably a very ghetto solution, maybe @cosmicLinux has a better idea!

neelcosmo:
Hi @gcanasherrera , @LisaGoh , I don't mean to butt in on this discussion but I have thought about a lot of the same questions that you both are having as I'll be working on implementing hi_class (I was swamped with other projects the past few weeks, so I haven't managed to start a PR for this, but I opened an issue earlier). Would it be possible to include me in your discussions?

LisaGoh:
Hi @neelcosmo yeah sure I'm cool with that! Anything that saves us the effort of doing things twice~

LisaGoh:
We've cleaned up the cloelib cosmology folder such that it now only contains the mochi_class_cosmology.py Protocol file, and everything else pertaining to the stable basis parametrisation has been moved to a separate repo!

cloelib/cosmology/mochi_class_cosmology.py
cosmicLinux: Yes, this is to load the input arrays when there passed on to the python wrapper. It's been used in the official mochi_class example scripts so I would recommend keeping it.

gcanasherrera:
I'm going to try one more time (I'm really sorry, but, I never surrender :D ):

- @mcataneo can we renegotiate renaming the python wrapper of mochi_class, please?

mcataneo:
Hi @gcanasherrera, sorry just saw this.

I’ve already briefly discussed this with @cosmicLinux, and we couldn’t identify a sufficiently strong motivation for changing the name at this stage.

More generally, could you give some concrete examples of use cases that cannot be handled cleanly via virtual environments? Traditionally, the hi_class developers have kept the name of the Python wrapper unchanged, since the standard class code is nested within hi_class.

From my perspective, virtual environments remain the most straightforward and robust way to work with multiple class-based extensions in parallel, ensuring that each one stays self-contained and avoids conflicts altogether.

mcataneo:
To use the renamed python wrapper 'mochi_classy' reinstall mochi_class by first doing a checkout to the newly created branch mochi_classy in the public repo https://github.com/mcataneo/mochi_class_public

gcanasherrera:
Thanks everyone! Also @mcataneo, cheers! 

- @LisaGoh could you check that th new mochi_class branch of mochi_class is compatible with this branch? 
- @mcataneo join the wave and put a cool picture in GitHub so you appear all cool and shiny in the all-contributors of the cloelib README
- @LisaGoh could you please add in the README that we have enabled mochi_class interface?
- @cosmicLinux can you please make sure that the way of outputting sigma8 makes sure that you passed z=0 during initialization? (check how we did it in CAMB).

gcanasherrera:
@all-contributors please add @LisaGoh for code, docs, ideas

gcanasherrera:
@all-contributors please add @cosmicLinux  for code, docs, ideas

LisaGoh:
> Thanks everyone! Also @mcataneo, cheers!
> 
> * @LisaGoh could you check that th new mochi_class branch of mochi_class is compatible with this branch?
> * @mcataneo join the wave and put a cool picture in GitHub so you appear all cool and shiny in the all-contributors of the cloelib README
> * @LisaGoh could you please add in the README that we have enabled mochi_class interface?
> * @cosmicLinux can you please make sure that the way of outputting sigma8 makes sure that you passed z=0 during initialization? (check how we did it in CAMB).

sorry I missed this, just updated the docs!

gcanasherrera:
@LisaGoh can you checkout the conflicts so we are ready to merge?

gcanasherrera:
Conflicts need to be solved.

Also @santiagocasas could you take of this, too?

gcanasherrera:
@LisaGoh (CC- @santiagocasas):

note how class is now interfacing in the main branch, where it requires an optional linear module to be passed to the non-linear module. This is to allow for cloelike interface. Please, follow the same approach in mochi_class

LisaGoh:
Ach sorry this took so long, somehow the notifications completely slipped through! I'll update the branch and push the commits now

LisaGoh:
Hmm it seems like now the input of `CAMBNonLinearPerturbations` and `CLASSNonLinearPerturbations` are not consistent, in that the `log10TAGN` parameter is missing in the input of `CLASSNonLinearPerturbations`, and `CLASSNonLinearPerturbations` requires an addition `hmcode_version` parameter as input. How will this be reconciled in the `EuclidLikelihood_photo_Cls.py` script?

gcanasherrera:
Hi @LisaGoh, we can address that in a follow up task. Could you please open an issue?

In the meantime, have been the other comments by @santiagocasas been addressed?

LisaGoh:
I haven't yet heard from @santiagocasas, but all the comments have been resolved!

gcanasherrera:
@santiagocasas can you give your thumbs up, whenever you have the time?

santiagocasas:
@gcanasherrera  @LisaGoh  really busy week unfortunately, but I will look into it in between trains, by end of the week, hopefully!

cloelib/cosmology/mochi_class_cosmology.py
santiagocasas: we probably need to find a better solution for this, to avoid I/O errors in the future when running this on HPC

santiagocasas: after careful consideration, this is quite dangerous, this is a global modification of numpy formatting, that activates as soon as someone imports this mochi_class module. While at the moment nothing is happening, it could have unwanted consequences in the future. For reasons of time, we can approve soon this PR, but please @LisaGoh  @cosmicLinux  open a bug/enhancement issue on this so that we remember to find a solution.

cloelib/cosmology/mochi_class_cosmology.py
```diff
@@ -0,0 +1,849 @@
...
+
+    c0 = SPEED_OF_LIGHT / 1000
+
+    def __init__(
+        self,
+        H0: float,
+        Omega_b0: float,
+        Omega_cdm0: float,
+        Omega_k0: float,
+        As: float,
```
santiagocasas, line 41: mochi class needs to accept here alpha_s, check hi_class implementtion for an example. This is required by cloelib's background protocol now.

cloelib/cosmology/mochi_class_cosmology.py
```diff
@@ -0,0 +1,849 @@
...
+        Return the matter density as a function of redshift.
+
+        Args:
+            zs (np.ndarray): Array of redshifts.
+
+        Returns:
+            np.ndarray: Matter density values.
+        """
+        return self.results.Om_m(zs)
+
```
santiagocasas, line 441: mochi_class needs an Omega_cb(zs): method returning the combined CDM+baryon density fractio, required by the CLASS protocol.

cloelib/cosmology/mochi_class_cosmology.py
```diff
@@ -0,0 +1,849 @@
...
+        background : dict
+            Dictionary of all background quantities at each time
+        """
+        return self.results.get_background()
+
+    @property
+    def rdrag(self) -> float:
+        """Sound horizon radius at last scattering in Mpc."""
+        return self.results.rs_drag()
+
```
santiagocasas, line 518: Background requires a z_star property. Could you add it here, matching hi_classBackground?

cloelib/cosmology/mochi_class_cosmology.py
```diff
@@ -0,0 +1,849 @@
...
+        elif hmcode_version is not None:
+            self.interface_args["CLASSparams"]["hmcode_version"] = hmcode_version
+        self.interface_args["CLASSparams"]["z_max_pk"] = np.max(self.z)
+        self.results = Class()
+        self.results.set(self.interface_args["CLASSparams"])
+        self.results.compute()
+        self.k = np.logspace(np.log10(1e-4), np.log10(self.kmax), 100)
+
+        try:
+            self.results.compute()
```
santiagocasas, line 766: why is there a duplicated compute here? the first one is outside the exception handler.

cloelib/cosmology/mochi_class_cosmology.py
```diff
@@ -0,0 +1,849 @@
...
+        self.interface_args["CLASSparams"]["output"] = "mPk, mTk"
+        self.interface_args["CLASSparams"]["P_k_max_1/Mpc"] = self.kmax
+        self.interface_args["CLASSparams"]["k_per_decade_for_bao"] = 70
+        self.interface_args["CLASSparams"]["k_per_decade_for_pk"] = 10
+        self.interface_args["CLASSparams"]["z_max_pk"] = np.max(self.z)
+        self.interface_args["CLASSparams"]["nonlinear_min_k_max"] = 50
+        self.interface_args["CLASSparams"]["hmcode_tol_sigma"] = 1e-8
+        self.interface_args["CLASSparams"]["non_linear"] = nonlinear_model
+        if background.mg_stable_basis_on:
+            self.interface_args["CLASSparams"]["non_linear"] = "none"
```
santiagocasas, line 747: why does a background switch change the behavior of a nonlinear interface? it should reject the nonlinear setting, but not return linear for a nonlinear regime.

cloelib/cosmology/mochi_class_cosmology.py
```diff
@@ -0,0 +1,849 @@
...
+        return D_z_k
+
+    def growth_rate(self) -> np.ndarray:
+        """
+        Calculate the growth rate f(z).
+
+        Returns:
+            (np.ndarray): Scale-independent growth rate f(z)
+        """
+        arr = [self.results.scale_independent_growth_factor_f(zi) for zi in self.z]  # type: ignore[union-attr]
```
santiagocasas: why a scale-independent growth rate? in MG, this is generally not the case. Either this needs a justification or some tests. Or big warnings to avoid problems and incompatibilities with D_z_k above.

santiagocasas (changes requested):
apologies for the delay @LisaGoh and thanks for your work!
I have added some comments in specific lines. The most important changes are regarding the Protocol requirements, these are easy to do. A few comments on the physics of the settings.

A few questions:

- There is no information in the README on how to install the mochi_classy wrapper. Also, it does not appear as optional dependency in the pyproject.toml, so the user does not really know how to install it. Can mochi_class be installed using pip?
- There are no tests at all for mochi_class, so the CI pipeline won't test this. A few set of examples can be seen in the hi_class implementation. However, this can only be done if the point 1 is resolved. If there is no easy way to install mochi_class, then there is no easy way to test it.

After the comments are resolved, please ping me so I can handle this quickly. Thanks!

cloelib/cosmology/mochi_class_cosmology.py
LisaGoh: This functionality hasn't been tested yet;  will switch it to the scale-dependent method instead to ensure consistency!

cloelib/cosmology/mochi_class_cosmology.py
LisaGoh: We do not plan to use mochiCLASS to calculate the nonlinear power spectrum for MG models, so right now it's hardcoded to only ever return the linear Pk, such that calculations with cloelike doesn't throw an error. It's probably not the most elegant method. What do you mean by 'reject the nonlinear setting'?

LisaGoh:
Thank you @santiagocasas for this! I've made changes to the script according to your comments. The main question I have left is what to do with the nonlinear implementation (it's mostly a placeholder that's forced to return the linear Pk for now, and will get overriden by React). 

Regarding your questions:

1. I've updated the `pyproject.toml` and `readme` files to include the mochi_class installation dependency. It will automatically clone the repo and install it. 
2. I've written up a simple test script (following the hi_class example), it passes for the LCDM implementation.

I hope these answers your queries! I'll also open another issue looking into the `np.set_printoptions` setting once this is merged.

LisaGoh:
Hmm the automatic pip installation of mochi_class is failing because of the gcc compiler. I don't think this is something we can control, so maybe we can't write tests for it as of now. Perhaps then I'll add a note in the readme on how to manually install mochi_class?

cloelib/cosmology/mochi_class_cosmology.py
santiagocasas: Yeah, I understand that basically none of the modified classy codes have a proper nonlinear Pk. However, instead of returning linear, when a nonlinear flag is passed, it would be "safer" to just reject the nonlinear flag or return a warning or something. However, as you say, we need to be careful with how this is handled in cloelike.
I think we can leave this now as is, and anyway I have plans to unify all the classy-related interfaces, because this is a recurring problem.

cloelib/cosmology/mochi_class_cosmology.py
```diff
@@ -658,7 +678,7 @@ def growth_rate(self) -> np.ndarray:
         np.ndarray
             Scale-independent growth rate f(z)
         """
-        arr = [self.results.scale_independent_growth_factor_f(zi) for zi in self.z]  # type: ignore[union-attr]
+        arr = [self.results.scale_dependent_growth_factor_f(1.0, zi) for zi in self.z]  # type: ignore[union-attr]
```
santiagocasas: so, this is still scale-independent but fixed at a particular scale of k=1.0 (Mpc or h/Mpc)?
This is also a problem that we have in hiclassy and mgclass, but we have at the moment k=0.1 or k=1.0. Could you instead of setting the number here, make it into a variable (such as k_fix) and explain it? So that at least we can track it back later on?

santiagocasas:
> Hmm the automatic pip installation of mochi_class is failing because of the gcc compiler. I don't think this is something we can control, so maybe we can't write tests for it as of now. Perhaps then I'll add a note in the readme on how to manually install mochi_class?

Thanks @LisaGoh !  
We had a similar problem with mgclass, so let me try to fix it with a patch as I did for the other one. I let you know during the day.

cloelib/cosmology/mochi_class_cosmology.py
LisaGoh: I've added a print warning for now!

cloelib/cosmology/mochi_class_cosmology.py
LisaGoh: We weren't using this functionality so  hadn't really considered this previously and simplistically copied over what was done in hi_class, but now after mulling it over I think we want to include the flexibility of scale dependence. I've added the wavenumber as an optional input, which defaults to 1 so as to not break the calculation of the Cl's downstream (when `need_rsd=True`).

santiagocasas:
@LisaGoh  I added patch to [cloe-org-environments ](https://github.com/cloe-org/cloe-org-environments) and to the .github/workflows/ci.yaml, so that mochiclass can be installed and tests can be run. Let's wait for the pipeline to pass here (I tested [locally](https://github.com/cloe-org/cloe-org-environments/wiki) with gh and act). If this works and this also does not break main and other repos, we can merge.
In general, the fix for mochiclass should be easy, I can try to write something in a PR in the github repo of mochiclass. The problem is just a non-standard Makefile requesting gcc-15.2. 
After that, we can remove the patch here and continue normally.
Thanks for the other fixes @LisaGoh !

LisaGoh:
Thank you! The CI seems to pass now, I think everything's up and working, and this PR can be merged @gcanasherrera @santiagocasas !

santiagocasas:
thank you @LisaGoh !  Glad that it works on your end as well!  I cannot merge yet, since I modified the `ci.yaml` just for this branch (due to the gcc issue). I will test the workflow on the main branch and once it passes there as well, I can then finally merge. I will do that tomorrow.

santiagocasas (approved):
@LisaGoh  needs some checks on my end to make sure the new workflows/ci.yaml doesn't break the main branch

santiagocasas:
@gcanasherrera I think you can now safely remove your changes suggestions or click on approve. I will merge after the update to the cloe-org-environments.

santiagocasas:
@gcanasherrera  since #566  was merged and passed, I did one final commit here to double check. This should pass and can be merged.

gcanasherrera:
Conflicts needs to be resolved

PR #355, opened 2025-12-01
gh-356: np.arange fix, now to n+1

caspervedder:
### ✅ Summary

COSEBIS range from 1 to Nmax, hence the correct indexing is np.arange(1,Nmax +1), the previous version had np.arange(1, Nmax), currently all works fine if you use Nmax = Nmax + 1, but this can easily lead to errors. 

### 🔄 Changes

See above


gcanasherrera:
@all-contributors please add @caspervedder for bugs

gcanasherrera (approved):

PR #368, opened 2026-01-06
gh-367: Implementation of EE2

### LINKED ISSUE #367: Implementation of EuclidEmulator2
## 📝 Issue Summary

Adding the possibility to compute the matter power spectrum with the Euclid Emulator 2 (EE2).

### ✅ Expected Outcome

- **Goal 1**: EE2 predictions can be obtained in cloelib

### 📌 Additional Notes

Will add a new cosmology file with the EE2 prediction for the boost in a new Perturbations class.

The Euclid Emulator 2 version that is expected to be most compatible is at https://github.com/PedroCarrilho/EuclidEmulator2/tree/pywrapper

PedroCarrilho:
### ✅ Summary

Implemented Euclid Emulator 2 for predicting the nonlinear matter power spectrum

### 🔄 Changes

- EE2_cosmology created with single class EE2NonLinearPerturbations that computes the nonlinear matter power spectrum, by applying a boost to the linear matter power spectrum that must be passed.
- Altered the behaviour of the extrapolator for low wavenumber, so that it smoothly approaches 1 instead of just setting it to 1. 
- Added some unit tests for EE2_cosmology

### 🛠 How to Test

To use it, just do

```
from cloelib.cosmology.EE2_cosmology import EE2NonLinearPerturbations

EE2_nonlinear = EE2NonLinearPerturbations(background=camb_instance, linearperturbations=camb_linear,
                                           redshifts=z)

nonlinear_pk_EE2 = EE2_nonlinear.matter_power_spectrum(z, ks)

```
assuming you have previously created the appropriate camb quantities, z and ks arrays. Can also be passed to C_ell calculations as usual.

### 📝 Documentation

### 🏗 Related Issues

Resolves #367 

Ratio of HMcode to EE2 P(k)
image

Same for GC C_ell
image

Same for WL

image

### 📌 Additional Notes

Made some particular choices for extrapolation and interpolation. In particular, I chose to output the wavenumber from the linear input instead of from EE2 output, since those are (by default) passed to the C_ell calculation and create some issues. I can explain this further, if necessary.

---


gcanasherrera:
Hi @PedroCarrilho, stupid, but euclidemu2 needs to be included in the CI.yaml

PedroCarrilho:
Hi @gcanasherrera , I added euclidemu2 to the CI, but that is not enough, as it needs GSL. There are then two options:
- Install gsl with conda, but that requires creating a conda environment, as far as I understand. This would change the entire CI file, I think.
- Install gsl with apt-get, but unclear if this would work, as I have no way of testing it (besides just putting it in the CI file and seeing what happens when I push it)

Which do you think is best?

santiagocasas:
hi @PedroCarrilho  and @gcanasherrera , I'll take a look at this between today and Friday! It should be doable.

santiagocasas:
Hi @gcanasherrera and @PedroCarrilho , I have analyzed this plus the whole thing about the cloe-org environments. And this is what we should do:

- The github `ci.yaml` should not have anything about our scientific stack (camb, classy, emulators...). These should all go in the pyproject,toml and pinned there to the version that works with tests and these should only be installed from there.
- The ci.yaml will then have a mamba/conda-forge environment and contain gsl. Then, cloelib will be installed with `pip install .[camb,classy,euclidemu2]" and that will suffice.
- I have fetched this branch by @PedroCarrilho  and ran the tests and it works. I would also pin euclidemu2==1.4.3, since it is the version with which the tests pass successfully. Pedro, since you are the maintainer of that pip package, you can always let us know if a requirement changes.

santiagocasas:
Now, one question @PedroCarrilho @gcanasherrera : 
- A) Should we go with a very dirty fix right now and put everything in this branch (I can do it), which means put here the environment file, change the toml, etc... so by merging this we have everything working?
- B) Create another PR, where I change the .toml (very minimally for now), create a new env.yaml and change the CI so that everything goes smoothly and THEN we merge this.

I would definitely go with B) because it is cleaner and lets us test the approach first, but if you are very stressed with time, we can do A) (not recommended). I don't think any other current PR depends on this or on changes in the CI, right @gcanasherrera  ?

santiagocasas:
Guys, since https://github.com/cloe-org/cloelib/pull/430 is working perfectly, this PR just needs : 
- revert the changes you did to the ci,yaml
- once #430 is merged to main `git merge main`
- commit and push, and it should work directly

gcanasherrera:
Hola @santiagocasas thanks a lot for this!

I would personally go for B), but I would also push the env to cloe-org-environments. That organisation is public so retrieving the raw is straightforward, I don't think you need any sort of KEY.

santiagocasas:
@gcanasherrera  okay, agreed!

santiagocasas:
Done @gcanasherrera  and @PedroCarrilho !  
Once https://github.com/cloe-org/cloelib/pull/430 is merged into main, I can take care of finishing this PR (or Pedro) by following the instructions above, which is quite easy.

PedroCarrilho:
Thanks @gcanasherrera and @santiagocasas ! I think this is mostly done and the tests pass, but I had to fudge something a bit. The new protocol requires the cb spectrum, which EE2 does not give you. I currently was just applying the EE2 boost to the cb linear, but this is just a place holder and a more correct thing may be possible, like the approximation in https://arxiv.org/pdf/2207.13677. I can test this a bit and do another commit tomorrow or leave it with the place-holder and open a different issue. Let me know if you have a preference (I would slightly prefer to do it as well as possible now).

PedroCarrilho:
Hi again @santiagocasas and @gcanasherrera,

I have now added a corrected version of P_cb that matches Bacco (but actually makes very little difference). Since this now appears to pass all tests, I suppose we can go ahead and merge it whenever appropriate in terms of version targets, but let me know if there is anything else that still needs changing.

CarmelitaCarbone:
Hi @PedroCarrilho so just to understand, you have implemented Pcb with BaccoEmu from @garico92 [here](https://github.com/cloe-org/cloelib/pull/411)  and from EuclidEmu2 following the prescription from [Castorina et al. 2015](https://arxiv.org/abs/1505.07148) and [Parimbelli et al. 2022](https://arxiv.org/pdf/2207.13677), right? 
The fact that the difference is really small is expected, as also explained in @GabrieleParimbelli  paper, but is still worth to implement both approaches. 
One question: about P_nu we have only the linear one from CAMB/CLASS, right?

PedroCarrilho:
> Hi @PedroCarrilho so just to understand, you have implemented Pcb with BaccoEmu from @garico92 [here](https://github.com/cloe-org/cloelib/pull/411) and from EuclidEmu2 following the prescription from [Castorina et al. 2015](https://arxiv.org/abs/1505.07148) and [Parimbelli et al. 2022](https://arxiv.org/pdf/2207.13677), right? The fact that the difference is really small is expected, as also explained in @GabrieleParimbelli paper, but is still worth to implement both approaches. One question: about P_nu we have only the linear one from CAMB/CLASS, right?

Hi @CarmelitaCarbone , let me explain further. I now use
Pcb_NL = Pcb_L + (Pmm_NL-Pmm_L)/f_cb^2
which is equivalent to Eq. 4.3 in [Castorina et al. 2015](https://arxiv.org/abs/1505.07148), but is not equivalent to Eq. 4.2 in [Parimbelli et al. 2022](https://arxiv.org/pdf/2207.13677), since that assumes Pcnu_NL to be different from Pcnu_L (and equal to sqrt(Pcb*Pnunu)). 

To clarify, what I was using before was instead
Pcb_NL=Pmm_NL/Pmm_L * Pcb_L
which is different, but the effect appears to be quite small (less than 0.1%) according to my tests.

Regarding P_nunu, we could get it from CLASS/CAMB, but we currently don't do that anywhere, and I am not sure it is necessary (I think we can write it using Pcb_L and Pmm_L).

CarmelitaCarbone:
Hi @PedroCarrilho, ok, I agree with your implementation, using Eq. 4.3 from Castorina et al. 2015. Thanks! 
About Pnunu_L and Pcnu_L, I wonder if they could be needed for galaxy-galaxy-lensing... but I've to think about...   

 

> > Hi @PedroCarrilho so just to understand, you have implemented Pcb with BaccoEmu from @garico92 [here](https://github.com/cloe-org/cloelib/pull/411) and from EuclidEmu2 following the prescription from [Castorina et al. 2015](https://arxiv.org/abs/1505.07148) and [Parimbelli et al. 2022](https://arxiv.org/pdf/2207.13677), right? The fact that the difference is really small is expected, as also explained in @GabrieleParimbelli paper, but is still worth to implement both approaches. One question: about P_nu we have only the linear one from CAMB/CLASS, right?
> 
> Hi @CarmelitaCarbone , let me explain further. I now use Pcb_NL = Pcb_L + (Pmm_NL-Pmm_L)/f_cb^2 which is equivalent to Eq. 4.3 in [Castorina et al. 2015](https://arxiv.org/abs/1505.07148), but is not equivalent to Eq. 4.2 in [Parimbelli et al. 2022](https://arxiv.org/pdf/2207.13677), since that assumes Pcnu_NL to be different from Pcnu_L (and equal to sqrt(Pcb*Pnunu)).
> 
> To clarify, what I was using before was instead Pcb_NL=Pmm_NL/Pmm_L * Pcb_L which is different, but the effect appears to be quite small (less than 0.1%) according to my tests.
> 
> Regarding P_nunu, we could get it from CLASS/CAMB, but we currently don't do that anywhere, and I am not sure it is necessary (I think we can write it using Pcb_L and Pmm_L).

cloelib/cosmology/EE2_cosmology.py
```diff
@@ -0,0 +1,229 @@
...
+    import euclidemu2
+
+    ee2 = euclidemu2.PyEuclidEmulator()
+
+except ImportError:
+    raise ImportError("EuclidEmulator2 could not be imported or initialised.")
+
+
+class EE2NonLinearPerturbations:
+    """Class for nonlinear perturbations using EE2, inheriting from Perturbations parent class."""
```
gcanasherrera: Same as before, we don't inherit, we make this compatible with the protocol.

gcanasherrera (changes requested):
Still pending:

- Update docs (both the docs/code_structure/perturbations.md) and the API adding this new module
- Update README
- Add EE2 to the .toml as additional dependency.

PedroCarrilho:
> Still pending:
> 
> * Update docs (both the docs/code_structure/perturbations.md) and the API adding this new module

Done. Also added `matter_power_spectrum_cb` as a required method, since it was missing. All units in the docs are shown as h/Mpc, which is probably not correct (at least in most cases). I can also fix that.
> * Update README

Done.
> * Add EE2 to the .toml as additional dependency.

Was already there(`euclidemu2 = ["euclidemu2==1.4.3"]`), unless you meant something else that I didn't understand.

@gcanasherrera , I think that is everything, but let me know if I missed something.

cloelib/cosmology/EE2_cosmology.py
PedroCarrilho: Fixed

gcanasherrera:
Indeed @PedroCarrilho, the toml wasn't flagged as modified files, but because Santi took care of this before. You just need to git pull from origin main to be able to pass CI. I changed yesterday the call to trapezoid.

PR #422, opened 2026-02-04
gh-268: including CMB lensing

### LINKED ISSUE #268: Implement CMB Lensing Tracer
## 📝 Issue Summary
This is quite a simple task, we need to implement a CMB lensing (preferentially in a way that can be extended to other CMB secondary observables too)  into the observables. This would support the use of cloelib for the CMBX measurements.

- [ ] **Goal 1**: Add a `observables/cmb.py` including classes that model the CMB secondary observables in the same way as other observables in cloelib
- [ ] **Goal 2**: include unit tests
- [ ] **Goal 3**: benchmark it against CCL and other codes like class.

### ✅ Expected Outcome
Being able to calculate theory vectors for 3x2pt photo + cross-correlations with cmb lensing.

### 🐞 Current Behavior (if applicable)
None, we do not have that implemented hehe

### 📌 Additional Notes
Simple, happy to take on this task, I just implemented this in firecrown too :) 

### 👥 Assignees
- @arthurmloureiro 
- @raphkou

raphkou:
### ✅ Summary

This PR allows computing CMB lensing power spectrum and cross-correlations with position and shear tracers.

### 🔄 Changes

- I added a new observable cmb.py which allows computing CMB lensing window function.
- I modified angular_two_point.py to compute CMB lensing power spectrum and cross-correlations. In particular, I modified the variable C_ell_out to handle the case where the tracers have different numbers of bins. This is the case of CMB lensing, which has only one redshift bin.
- I modified several cosmology files (camb, class, jax) to compute z_star which is required to compute the CMB lensing window function.
- I added a few unit tests in test_cmb.py, allowing to check the computation of the CMB lensing window function, power spectrum, and cross-correlations.

### 🛠 How to Test

I added the notebook cmb.ipynb in the playground (see PR https://github.com/cloe-org/playground/pull/116), which shows how to compute the CMB lensing power spectrum and cross-correlations, and performs a benchmark against CAMB and CCL.

### 📝 Documentation

- [x] This PR updates documentation

### 🏗 Related Issues

Resolves #268.

### 📌 Additional Notes

Pseudo-cl computation not implemented yet.

---


cloelib/cosmology/derived_cosmology.py
```diff
@@ -79,3 +79,23 @@ def rdrag_fitting_function(background, neff=3.046):
...
+
+    Assumes a cosmology-independent z_star.
+
+    Parameters
+    ----------
+    background: Background
+        Background class containing cosmology
+    neff: float
+        Effective number of neutrinos.
+
```
arthurmloureiro: `neff` is not being used. I suggest we remove it.

(Also the `Neff` value we are fixing is `3.044` 😄 )

cloelib/observables/cmb.py
```diff
@@ -0,0 +1,84 @@
...
+            Redshift at which window kernel is being evaluated
+
+        Returns
+        -------
+        window: np.ndarray
+        """
+        Omega_m0 = self.background.Omega_m(0.0)
+        factor = (
+            3
+            / 2
```
arthurmloureiro, line 73: I think you can put these in the same line, `3 / 2`. 

Note that in python 3.0 if you divide two integers, you get an integer back. This is not a problem here because the other quantities being multiplied will be floats. But could cause some `jit` issues with ambiguous typing in `jax`.

cloelib/summary_statistics/angular_two_point.py
```diff
@@ -219,10 +226,23 @@ def she_she_rule(C, i, j):
...
+            block = C[:, i - 1, j - 1]
+            return {("CMBL", "SHE", i, j): np.stack([block, np.zeros_like(block)])}
+
         tracer_rules = {
             (PositionsTracer, PositionsTracer): pos_pos_rule,
             (PositionsTracer, ShearTracer): pos_she_rule,
             (ShearTracer, ShearTracer): she_she_rule,
+            (CMBLensingTracer, PositionsTracer): cmbl_pos_rule,
+            (CMBLensingTracer, ShearTracer): cmbl_she_rule,
+            (CMBLensingTracer, CMBLensingTracer): cmbl_cmbl_rule,
```
arthurmloureiro, line 356: It seems from testing with your playground notebook that there's no symmetry in the calling. For example, the following will break:
```python
AngularTwoPoint(tracer_pos, tracer_cmblens_cross)
```
with the following error:
```python
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[25], line 3
      1 plt.figure()
      2 for i in range(6):
----> 3     ells = cells['CMBL', 'POS', 1, i+1].ell
      4     plt.loglog(ells, ells*(2*ells+1)*cells['CMBL', 'POS', 1, i+1], label = '$n_{{{}}}$'.format(i+1), linewidth=2, color=sns.color_palette("rocket", len(nz_example.keys()))[i-1])
      5 plt.xlabel(r'$\ell$')

KeyError: ('CMBL', 'POS', 1, 2)
```

Is this a feature or a bug? 😄

arthurmloureiro (changes requested):
Excellent stuff @raphkou !

Just a few questions and comments!

cloelib/cosmology/derived_cosmology.py
raphkou: This was copying `rdrag_fitting_function(background, neff=3.046)` in case we implemented a fitting function that depended on `neff`, but I'm happy to remove it since it is not the case at the moment.

cloelib/observables/cmb.py
raphkou: I'm happy to do that as well, but this is really copying `get_window_lensing` from `photo.py`. Should we change that there as well then?

cloelib/summary_statistics/angular_two_point.py
raphkou: There is actually some kind of symmetry because in your example, the key `('CMBL', 'POS', 2, 1)` would be present. But I agree it's not really logical since the 2 refers to POS. We have the choice whether your case should give `('CMBL', 'POS', 1, 2)` or `('POS', 'CMBL', 2, 1)`. I believe the former option would be more consistent with what is already implemented for the main probes, as `POS` always come before `SHE` for instance. If you agree with that, I will modify the code accordingly.

gcanasherrera: In the SGS a decision in terms of order was done for 3x2pt (POS and WL). I encourage to do the same here.

cloelib/cosmology/cosmology.py
```diff
@@ -126,6 +126,11 @@ def rdrag(self) -> float:
         """Sound horizon radius at last scattering in Mpc."""
         ...
 
+    @property
```
gcanasherrera, line 133: @raphkou @arthurmloureiro @cloe-org/cloe-maintainers 

Do we actually need this to be part of the background protocol? As far as I understand, it is a similar case as r_drag... can we take it to derived_cosmology? 

Forcing it to be part of the protocol would make everyone (even if you are coding GW in the future) to drag this along.

cloelib/cosmology/derived_cosmology.py
raphkou: @arthurmloureiro I removed it in the latest commit.

cloelib/summary_statistics/angular_two_point.py
raphkou: @arthurmloureiro @gcanasherrera I updated this in the latest commit. get_Cl will always return `('CMBL', 'POS', i, j)` in that order, and i will refer to 'CMBL' and j to 'POS'. It should also be symmetric in the calling.

cloelib/cosmology/cosmology.py
raphkou: @gcanasherrera Thanks a lot for the comment. However I'm not completely sure how to do that. In the new observable cmb.py, I use `z_star = self.background.z_star` which is then required to compute the window function. In this PR, `z_star` is now available in camb_cosmology.py, camb_cosmology.py and jax_cosmology.py. I can remove it from cosmology.py, but cmb.py will still assume that its background object knows what `z_star` is.

I am not sure I can take it to derived_cosmology.py, because this is a quantity that is computed by CAMB/CLASS (for jax we fix it to 1090).

arthurmloureiro: I think it may make sense to be a part of derived cosmology but I have not yet played with it to suggest an implementation

gcanasherrera:
Hey @raphkou , I was reviewing this more in detail, and I guess it is fine the way that it is, but, could we please update the docs and make sure that we are compatible with `numpy`>2 before merging? Thanks

raphkou:
Hi @gcanasherrera, thanks for reviewing this PR. I have updated the code so that it is indeed compatible with numpy>2. It seems to be alright now.

gcanasherrera:
For Guada to remember: this is being well taken care of, just on hold due to holidays.

gcanasherrera (approved):
Many thanks @raphkou

gcanasherrera:
@all-contributors please add @raphkou for code, ideas, docs

PR #423, opened 2026-02-05
gh-403: compatibility photo class

### LINKED ISSUE #403: Compatibility problem in photo.py when using CLASS perturbations
### ⚠️🐛 Bug Summary

Using the photo.ipynb notebook, but replacing CAMB by CLASS linear perturbations, I get errors as shown in the screenshots. It seems this could be resolved by making sure that the arguments given to CLASS functions are always arrays, which means replacing

`Omega_m0 = self.background.Omega_m(0.0) `

in photo.py by 

`Omega_m0 = self.background.Omega_m(np.array([0.0]))[0])`

I have the newest CLASS version 3.3.4 and @zsirap has encountered the same issue. A second compatibility issue arises due to the use of self.perturbations.k (which currently is defined for CAMB but not for CLASS) in lines 86-88 of photo.py:

```
Dz = self.perturbations.growth_factor(
            self.perturbations.z, self.perturbations.k
        )[:, 1]
```

### 🔗 Optional: Additional Context

Image

Image

zsirap:
### 🏗 Related Issues

Resolves #403 


cloelib/cosmology/class_cosmology.py
```diff
@@ -260,7 +264,11 @@ def Omega_b(self, zs: np.ndarray) -> np.ndarray:
         Returns:
             (np.ndarray): Matter density values.
         """
-        return np.array([self.results.Om_b(z) for z in zs])
+        try:
+            Omegab_0 = np.array([self.results.Om_b(z) for z in zs])
```
gcanasherrera: why Omegab_0? where is the 0 for? This is redshift dependent

cloelib/cosmology/class_cosmology.py
```diff
@@ -290,6 +298,7 @@ def __init__(self, background: Background, redshifts: np.ndarray):
         self.results = Class()
         self.results.set(self.interface_args["CLASSparams"])
         self.results.compute()
+        self.k = np.logspace(np.log10(1e-4), np.log10(self.kmax), 100)
```
gcanasherrera, line 305: Why are we defining the self.k here? This is not part of the issue

cloelib/cosmology/class_cosmology.py
```diff
@@ -248,7 +248,11 @@ def Omega_m(self, zs: np.ndarray) -> np.ndarray:
         Returns:
             (np.ndarray): Matter density values.
         """
-        return np.array([self.results.Om_m(z) for z in zs])
+        try:
+            Omegam_0 = np.array([self.results.Om_m(z) for z in zs])
```
gcanasherrera: why Omegam_0? where is the 0 for? This is redshift dependent

gcanasherrera (changes requested):
Few comments, worth @arthurmloureiro checking on this too

cloelib/cosmology/class_cosmology.py
zsirap: yes you are right .. my bad .. left it _0 because it was meant to solve our problem at z  =0 but it still works at any redshift so will change the notation only to Omegam

cloelib/cosmology/class_cosmology.py
zsirap: yes it is ... photo.py use perturbations.k and it fails in the case of CLASS 

you already recommended it 

> For the self.perturbations.k, simply save it as an attribute in the classy interface as we do in CAMB, but do not impose it at the level of the protocol.

gcanasherrera (approved):

gcanasherrera:
Looks fine on my side but I will let @arthurmloureiro explore as well.

gcanasherrera (approved):

gcanasherrera:
@arthurmloureiro seal of approval on this?

arthurmloureiro:
@gcanasherrera @zsirap on it :)

I checked and this bug already appears in classy 3.3.1 (I think it's the one we have in the CI, before the restructuring of classy)

cloelib/cosmology/class_cosmology.py
```diff
@@ -248,7 +248,11 @@ def Omega_m(self, zs: np.ndarray) -> np.ndarray:
         Returns:
             (np.ndarray): Matter density values.
         """
-        return np.array([self.results.Om_m(z) for z in zs])
+        try:
+            Omegam = np.array([self.results.Om_m(z) for z in zs])
+        except TypeError:
+            Omegam = self.results.Om_m(zs)
+        return Omegam
```
arthurmloureiro: @zsirap @NastassiaG I think the easiest solution to this problem is getting rid of the iteration here:
https://github.com/cloe-org/cloelib/blob/27a54a2d19b24bf22bc5725e468ebeb04bcd094a/cloelib/cosmology/class_cosmology.py#L263
and simply change that to:
```python
return self.results.Om_m(zs) 
```

This function in class will work with both arrays and floats:
Image

I do not know why it was doing the iteration over values of `zs` in the first place haha

cloelib/cosmology/class_cosmology.py
```diff
@@ -260,7 +264,11 @@ def Omega_b(self, zs: np.ndarray) -> np.ndarray:
         Returns:
             (np.ndarray): Matter density values.
         """
-        return np.array([self.results.Om_b(z) for z in zs])
+        try:
+            Omegab = np.array([self.results.Om_b(z) for z in zs])
+        except TypeError:
+            Omegab = self.results.Om_b(zs)
+        return Omegab
```
arthurmloureiro: Same here: 
```python
return self.results.Om_b(zs)
```
Works with floats and arrays:
Image

cloelib/cosmology/class_cosmology.py
arthurmloureiro: I am wondering if this hardcoded `kmin` won't come back to haunt us later haha

arthurmloureiro (changes requested):
@zsirap I think there's a simpler way of fixing this.

It's also worth noting you need to rebase with the main branch :)

cloelib/cosmology/class_cosmology.py
zsirap: I think they put the iteration because the function in CLASS used to fail with an array ... maybe they change it in the new version ...  so in the example you are showing, you have removed the iteration ... right ?

zsirap: > you have removed the iteration ... right ?

most probably you did .. so could you @arthurmloureiro check the other functions if they fail or pass without the iteration such as "angular_diameter_distance" ? because I think they used to fail as well .. if they don't anymore then we should go through class_cosmology.py and remove all unnecessary iterations ...

anyway I will later by all means check them if you already hadn't and report here ...

arthurmloureiro: I can check later, but I think this should be another issue/PR (following the 1 issue = 1 PR).

I checked it with classy 3.3.1 which is the one supported in the cloelib CI, so seems like we are safe 😄

zsirap: > I checked it with classy 3.3.1 which is the one supported in the cloelib CI, so seems like we are safe

ok so tomorrow will modify it accordingly and remove the iteration ...

zsirap: done @arthurmloureiro

arthurmloureiro (approved):
Awesome!

gcanasherrera (approved):

gcanasherrera:
@all-contributors please add @NastassiaG for code, bug, ideas

PR #430, opened 2026-02-13
gh-429: Updated CI pipeline and pyproject.toml minimally

### LINKED ISSUE #429: Improve and fix CI pipeline for tests of external codes
## 📝 Issue Summary

External codes like euclidemu2, need numerical/scientific libraries like gsl, requiring a CI pipeline with a mamba environment that is reproducible and robust.

### ✅ Expected Outcome

- **Goal 1**: Mamba environment installation in the CI pipeline
- **Goal 2**: All current tests passing
-  **Goal 3**: Tests for branch 367-euclidemu2 #368 should pass 

### 📌 Additional Notes

This should be merged before [368](https://github.com/cloe-org/cloelib/pull/368)

Following similar steps as for cloelike in : https://github.com/cloe-org/cloelike/pull/102/changes#top

santiagocasas:
### ✅ Summary

This PR modernizes the test CI workflow to use a dedicated Mamba environment file and installs backend dependencies through `pyproject.toml` extras instead of hardcoded CI package installs.
It also adds support for `euclidemu2` in project extras and introduces a repository-local environment spec for reproducible CI test setup.

### 🔄 Changes

- Reworked `.github/workflows/ci.yaml` test job to:
  - create/use Conda env from `environments/cloelib-env.yaml`
  - install package + extras via pip (`.[camb,classy,euclidemu2]`)
  - run tests with `python -m pytest`
- Added `environments/cloelib-env.yaml` as the local CI environment definition
- Updated `pyproject.toml` optional dependencies:
  - pinned `classy==3.3.1` (temporary compatibility pin)
  - added `euclidemu2==1.4.3`
  - pinned `camb==1.6.5`
- Kept build validation aligned with packaging flow (`python -m build`)

### 🛠 How to Test

```bash
# from repo root
mamba env create -f environments/cloelib-env.yaml

mamba run -n cloelib-env python -m pip install --upgrade pip
mamba run -n cloelib-env python -m pip install ".[camb,classy,euclidemu2]"

mamba run -n cloelib-env python -m pytest
mamba run --live-stream -n cloelib-env pre-commit run --all-files --show-diff-on-failure
mamba run -n cloelib-env python -m build
```

### 📝 Documentation

### 🏗 Related Issues

Resolves #429

N/A

### 📌 Additional Notes

- This is a temporary in-repo env-file approach (`environments/cloelib-env.yaml`) for simplicity.
- Follow-up can switch CI to use the canonical `cloe-org-environments` repository directly.
- `classy==3.3.1` is pinned due to test compatibility drift observed with newer `classy` versions.

---


.github/workflows/ci.yaml
```diff
@@ -1,68 +1,63 @@
 name: CI
-
 on:
   push:
   pull_request:
     branches:
       - main
-
+env:
+  ENV_NAME: cloelib-env
```
gcanasherrera: I think you can take this to cloe-org-environments, really.

pyproject.toml
```diff
@@ -32,10 +32,11 @@ dependencies = [
 dynamic = ["version"]
 
 [project.optional-dependencies]
-camb = ["camb"]
-classy = ["classy"]
+camb = ["camb==1.6.5"]
+classy = ["classy==3.3.1"]
```
gcanasherrera, line 36: I'm really confused about all the classy thing going on. @arthurmloureiro @santiagocasas can you please tell me the differences between 3.3.1 and 3.3.4, and why we can't go to 3.3.4 again? Thanks!!

environments/cloelib-env.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+  # --- Plotting / UX helpers ---
+  - matplotlib
+
+  # -----------------------------
+  # pip-only dependencies
+  # -----------------------------
+  # Euclid- and project-specific packages not available via conda
+  - pip:
+      - euclidlib==2026.1
```
gcanasherrera: Just pinging @AndreaPezzotta and @chiaramoretti to know that this will need to be updated after the new release of euclidlib with spectro!

environments/cloelib-env.yaml
```diff
@@ -0,0 +1,64 @@
+# =====================================================================
```
gcanasherrera: This can go to cloe-org-environments really (I know, I'm super annoying :D )

gcanasherrera (changes requested):
You're great Santi, thanks a lot!

Few comments.

pyproject.toml
santiagocasas: hi @gcanasherrera  yeah this seems to be a slight pain in the neck. Apart from other more "physical" reasons, the new API of classy==3.3.4 changed a few attributes, especifically these:
- with classy 3.3.4.0, CLASSBackground.N_eff and CLASSBackground.rdrag return classy._classy.CallableFloat objects (not plain float).
- In tests/test_class_cosmology.py:65, the attribute test builds attributes_found by excluding anything where callable(value) is true.
- CallableFloat is callable, so N_eff and rdrag get filtered out, and this assertion fails at tests/ test_class_cosmology.py:67.

In other words it's incompatible with the Protocol. 
I would keep the pin to 3.3.1 until we analyze this further. Maybe there are other reasons that were exposed in other recent PRs (such as the Weyl one).

environments/cloelib-env.yaml
santiagocasas: Sure! Actually, @gcanasherrera  I think euclidlib should go with the rest of the scientific stack in the dependencies in the toml. I left this here, to have a minimal modification only. 
But someone, for instance, using `uv` and not mamba (trusting its local env) and then installing cloelib through `pip install .` would only realize they are missing `euclidlib` at runtime. Which is quite dangerous.
I would put it in the .toml as a dependency and pin it there to whatever you prefer. Or remove the pin if you think you will be doing releases very often and you don't think tests will break.

.github/workflows/ci.yaml
santiagocasas: okay!

environments/cloelib-env.yaml
santiagocasas: okay, if you insist... and knowing that cloe-org-environments is public, I can put it there.
I would have to add one more step which is git pulling this env file and running it.
You should do the same for cloelike, as well.
But it's indeed the best solution and we should have just very few centralized env files (ideally just one for all platforms).

pyproject.toml
gcanasherrera: @santiagocasas thanks a lot for this, very useful. Can you open a follow up issue for us to explore this too?

environments/cloelib-env.yaml
gcanasherrera: cloelib is not dependent on euclidlib for anything, so none would really need it unless that you want to read data, so I think it should not be part of the toml.

pyproject.toml
gcanasherrera: @copilot could you please take Santiago's comment about classy and create an issue with our issue template to do it in the future?

gcanasherrera: Aha, copilot worked for this, so, @santiagocasas I will merge this PR.

gcanasherrera (approved):

gcanasherrera:
@all-contributors please add @santiagocasas for maintenance

environments/cloelib-env.yaml
santiagocasas: but if it's not part of the toml, then it shouldn't be part of the conda environment, neither

gcanasherrera: cloe-org-environments is to run all cloe-org, and playground and blindspot need it. This is why is in the env.

PR #511, opened 2026-05-01
gh-509: COSEBI dictionary and indexing

### LINKED ISSUE #509: COSEBI dictionary and indexing
### ⚠️🐛 Bug Summary

The W_ells are produced as a dict, but get_cosebi demands an array. This can lead to some awkward problems.

caspervedder:
### ✅ Summary

Very small indexing fix for the cosebis

### 🔄 Changes

get_cosebis now accepts the output of get_well, which is a dictionary, ns is turned from the cosebi mode to the array index by doing -1 
- Feature 1
- Feature 2
- Bug fix 1
- Other improvements

### 📝 Documentation

### 🏗 Related Issues

Resolves #IssueNumber

### 📌 Additional Notes

---


gcanasherrera:
Hey @caspervedder, can I proceed with this?

cloelib/summary_statistics/angular_two_point.py
gcanasherrera, line 185: Modify this for consistency too

gcanasherrera (changes requested):

gcanasherrera:
@all-contributors please add @caspervedder for bug, ideas

cloelib/summary_statistics/angular_two_point.py
caspervedder: You mean the argument descriptions? Will fix!

caspervedder:
> Hey @caspervedder, can I proceed with this?

Yes!

cloelib/summary_statistics/angular_two_point.py
gcanasherrera: If below, in the class, you will pass the dict, then, it makes sense that we pass the dict here too, no?

cloelib/summary_statistics/angular_two_point.py
```diff
@@ -162,7 +171,7 @@ def compute_cosebi(w_n):
             bb = np.sum(ells * cl_bb * w_n * weights) / (2 * np.pi)
             return ee, bb
 
-        ee_vals, bb_vals = jax.vmap(compute_cosebi)(w_ell[ns])
+        ee_vals, bb_vals = jax.vmap(compute_cosebi)(w_ell)
         arr = np.zeros((2, 2, ns.shape[0]), dtype=np.float64)
         arr = arr.at[0, 0, :].set(ee_vals)
```
gcanasherrera: would be good, although redundant, but you never know, to update the docs to make sure that the user understand that the element `0` corresponds to `well` for n = 1

gcanasherrera (changes requested):

cloelib/summary_statistics/angular_two_point.py
caspervedder: You mean specifying that in the output cosebi_E[0] corresponds to ns[0]? As the whole indexing is internal, we just turn w_ell into an array and then turn those into cosebis. In principle, the user just submits an array with n's (e.g. [2,3,4]) and gets the corresponding cosebis formatted like the returned ns.

gcanasherrera: I'm lost. You're keeping the input the dictionary, right?

So, effectively, the returned objects is not a dictionary but an array. So, I thought potentially users can get confused...?

caspervedder: The function takes the W_ell dict and the ns array, we can also make this only the W_ells, but this is how it is now. The n's can be any subset of the n's in W_ell. The returned cosebi's will simply be related to the (sorted) ns, which the function returns. Hence, the ordening of the cosebi's is always given by the n's (although sorted). Thus, cosebi[0] is the mode given by ns[0]. 

So basically, the user just determines the ordening themselves with the ns array. But then indeed, for an array starting from 1 it will mean element 0 is mode 1, but in principle the more general statement would be element[0] is mode ns[0]. 

I can write this a bit more cleanly in the docs

caspervedder: Basically, this way, it allows for any obscure combination of modes someone would want to query. For example, someone very interested in only mode 5 and 6 and not so much in 1-4 (which I guess woud be quite odd)

gcanasherrera: Just a quick line in the docs would do.

caspervedder: Done! Please check if you agree with this :)

cloelib/auxiliary/cosebi_helpers.py
```diff
@@ -221,6 +221,35 @@ def tm(n, t, tmin, nn, coeff_j):
     )
 
 
+def _tm_fast(n, thetagrid, tmin, nn, coeff_j):
```
caspervedder, line 224: Very nice, tested whether it matches my original tm (which we validated with Robert) and it does, and it seems to speed  up this function by a factor 3-6 for the n's that I checked!

cloelib/auxiliary/cosebi_helpers.py
```diff
@@ -221,6 +221,35 @@ def tm(n, t, tmin, nn, coeff_j):
...
+    # Log-polynomial: evaluate in mpmath via Horner's method (mp.polyval) to
+    # preserve precision against cancellation, then cast to float64 in one pass.
+    # mp.polyval expects coefficients highest-degree first.
+    dnm_coeffs_highfirst = [dnm(n, m, nn, coeff_j) for m in range(n, -1, -1)]
+    result += np.array([float(mp.polyval(dnm_coeffs_highfirst, zi)) for zi in z_np])
+
+    return result
+
+
 def get_W_ell(thetagrid, Nmax, ells, N_thread):
```
caspervedder, line 253: Do we want tomo bin functionality already at this level? Thus that is it would be possible to create tomo W_ells in this function, which you can plug in to the get_cosebis without any formatting in an efficient manner. If I understand correctly, now it is needed to do this formatting yourself. 

In the context of cloelike: for the 2pcf, one gives cloelike a bunch of keyed scale cuts, which are then applied. We could make it such that get_W_ells can accept such a dict. 

Two options for cloelike depending on this choice: 1) we pass it a scale cut dict, It then precomputes the W_ells for these and this dict is then forwarded in get_cosebis. 2) we simply omit the option to pass scale cuts and make the user precompute the W_ells themselves. This then can be one of the two options we allow for.

cloelib/summary_statistics/angular_two_point.py
```diff
@@ -117,6 +117,57 @@ def _growth_rate_on_grid(perturbations, zs_target):
...
+        if kernel_dict is None:
+            raise KeyError(
+                f"No w_ell kernel found for bin pair ({i}, {j}). "
+                "Provide either a (i,j)-keyed entry or a None fallback."
+            )
+    else:
+        # Global: the dict itself is the kernel dict
+        kernel_dict = w_ell
+
+    kernel_array = np.stack([np.asarray(kernel_dict[int(n)]) for n in ns], axis=0)
```
caspervedder, line 179: Maybe nice to throw an error if n in ns is not in kernel_dict? Although I suspect the standard python error that will be thrown will be fairly self explanatory, you decide!

cloelib/auxiliary/cosebi_helpers.py
caspervedder, line 224: Do we want to keep this _internal though? I can imagine quite some cases where people want to query these themselves? (but then I guess they can also import it explicitly, so you decide!)

cloelib/auxiliary/cosebi_helpers.py
caspervedder: Something along these lines, but then more efficient, as many entries will be the same. This will get a bit verbose though, due to the many print statements in the get_w_ell...  

nmax = 2 # 2 because I do not have all day

theta_ranges = {
    ("SHE", "SHE", i, j): (0.5, 300.0)
    for i in range(6)
    for j in range(i, 6)
}

well = {}

for key, (th_min, th_max) in theta_ranges.items():
    theta_integration = (
        np.geomspace(th_min, th_max, int(1e5)) * np.pi / 10800
    )
    well[key] = get_W_ell(
        theta_integration,
        nmax,
        ells_integration,
        N_thread,
    )

cloelib/summary_statistics/angular_two_point.py
```diff
@@ -117,6 +117,57 @@ def _growth_rate_on_grid(perturbations, zs_target):
...
+      ``"metadata"``).  The kernel for ``bin_key`` is looked up first; if
+      the pair is absent the ``(j, i)`` transpose is tried; if still absent
+      the global fallback (key ``None``) is used.
+
+    Parameters
+    ----------
+    w_ell : dict
+        Either a global kernel dict or a per-bin dict of kernel dicts.
+    bin_key : tuple
+        Full cells key ``('SHE', 'SHE', i, j)``; only the bin indices
```
caspervedder, line 153: If I key them like (SHE,SHE, i, j), it throws an error, but if I key them with only (i,j) it does not. The error message is clear, but the docs are a bit confusing I think. I think keying them as (SHE,SHE, i, j) is most consistent with the rest of cloelib, but you are the expert!

cloelib/summary_statistics/angular_two_point.py
```diff
@@ -117,6 +117,57 @@ def _growth_rate_on_grid(perturbations, zs_target):
...
+    # Detect per-bin layout: values are dicts (not arrays)
+    first_val = next(v for k, v in w_ell.items() if k != "metadata")
+    if isinstance(first_val, dict):
+        # Per-bin: try (i,j), then (j,i), then global fallback None
+        kernel_dict = w_ell.get((i, j)) or w_ell.get((j, i)) or w_ell.get(None)
+        if kernel_dict is None:
+            raise KeyError(
+                f"No w_ell kernel found for bin pair ({i}, {j}). "
+                "Provide either a (i,j)-keyed entry or a None fallback."
+            )
```
caspervedder, line 174: See comment above!

caspervedder:
@gcanasherrera I had a look at it, very nice :)! See mainly my comments on the w_ell dict keys and whether we want to have tomo functionality already in the get_w_ell function!

cloelib/auxiliary/cosebi_helpers.py
gcanasherrera: I am not happy about returning a dictionary of dictionaries, you can always construct that yourself, it's more maintenance. 

I would support the user to do exactly what you did themselves.

cloelib/summary_statistics/angular_two_point.py
gcanasherrera: I think it will be explanatory because it will complain about the n-bin

cloelib/auxiliary/cosebi_helpers.py
gcanasherrera: I wanted to further study how to make them faster, therefore the private methods for the moment.

cloelib/summary_statistics/angular_two_point.py
gcanasherrera: The thing is that COSEBIs will always be SHE SHE, no? That's why I was using only (i, j)....

cloelib/summary_statistics/angular_two_point.py
gcanasherrera: I will fix the docs

gcanasherrera:
@all-contributors please add @caspervedder for review

gcanasherrera:
Merging!!

PR #535, opened 2026-05-15
gh-510: Method sigma8_0 failing on repeated calls for HMemu perturbations

### LINKED ISSUE #510: Method sigma8_0 failing on repeated calls for HMemu perturbations
### ⚠️🐛 Bug Summary

The method self.sigma8_0() for class HMemuNonLinearPerturbations contains the following lines:

```
self.sigma8_0, _ = HM2020_emu.get_sigma8(**self.params_hm_emu)
return self.sigma8_0[0]
```

This only works fine when self.sigma8_0() is called the first time. When called a second time, self.sigma8_0 has been overwritten by an array and self.sigma8_0() leads to an error, see screenshot. Moreover, the function self.sigma8_0() is currently missing in the _linear_ perturbations class, HMemuLinearPerturbations. 

I suggest to:
- Replace self.sigma8_0 by self.sigma8 in the two lines above, to avoid overwriting the method by an array.
- Copy the (corrected) implementation for self.sigma8_0() as well into the linear perturbations class. 

### 🔗 Optional: Additional Context

Image

### LINKED ISSUE #575: Mismatch in length of growth rate array when calculating the RSD window function
### ⚠️🐛 Bug Summary

I have encountered the following error while testing the `include_rsd = True` option for [gh-139](https://github.com/cloe-org/cloelike/pull/140) in cloelike: 

`ValueError: operands could not be broadcast together with shapes (1,100) (1,101)`

When calculating the RSD window function, the length of the growth rate array `f` is 1 element too large compared to the `H` array, see also attached screenshot. The error seems to arise from the fact that the Hmemu perturbations class adds `z=0 `to the `self.z` array (if not already included in `redshifts` when initialising the perturbations object), resulting in the growth rate getting one extra entry. More precisely, the following seems to happen:

1. In `angular_two_point.py`, the function `_growth_rate_on_grid` is used to calculate `f`
2. The function `_growth_rate_on_grid` calls the function `perturbations.growth_rate()`
3. For the (linear or non-linear) HMemu perturbations class, the function `self.growth_rate()` evaluates the growth rate on the `self.z` grid.
4. `self.z` has been created through `self.z = ensure_z_zero_included(redshifts[redshifts <= redshift_max])`. This adds one extra element if zero was not initially included (which is the case for the array `myz = np.linspace(1e-4, 3.0, 100)` used in the playground example notebooks as well as in the unit tests). 
5. The `f` array then has length `len(redshifts)+1`, leading to the error message.

I’m not sure what the best fix to this is, options could be the following:
1. Modify the function `_growth_rate_on_grid()` in `angular_two_point.py`, e.g. such that it compares indices of `zs_target` and `self.perturbations.z`, and picks out the indices of `f_raw` accordingly.
2. Modify the function `growth_rate()` within the HMemu perturbations classes (and check if similar modifications are necessary for other perturbations classes).

### 🔗 Optional: Additional Context

Image

NastassiaG:
### ✅ Summary

This PR modifies the `self.sigma8_0()` method in `HMemuNonLinearPerturbations` to avoid failure on repeated calls. It is also adding the same method to `HMemuLinearPerturbations`, where it is currently missing.

### 🔄 Changes

- `HMemuNonLinearPerturbations`: Renaming the instance attribute `self.sigma8_0` assigned inside the method `self.sigma8_0()` to `self.sigma8`; this avoids the method being shadowed by an array (causing failure on repeated calls).
- `HMemuLinearPerturbations`: Implementing the missing `self.sigma8_0()` method by reusing the updated implementation from the nonlinear perturbations class.

### 🛠 How to Test

1. Create instances of both `HMemuLinearPerturbations` and `HMemuNonLinearPerturbations`.
2. Verify that the `sigma8_0()` method exists for both instances. 
3. Verify that, upon repeated calls, it yields the expected result with no exceptions raised.

See also screenshots below. 

### 📝 Documentation

### 🏗 Related Issues

Resolves #510 and #575

**Before implementing changes:**

Bildschirmfoto 2026-05-01 um 15 18 45

**After implementing changes:**

Bildschirmfoto 2026-05-15 um 09 22 46


gcanasherrera:
Hi @NastassiaG,

I was unable to reproduce your original reported bug, really. I've been running chains with this implementation for months now. Before we proceed with the review, could you specify a bit more what the issue could potentially be?

NastassiaG:
Hi @gcanasherrera sorry for my very late reply. I've recently made a clean installation in a new environment, and still see the same error: `nonlinear_perturbations_emu.sigma8_0()` works the first time it is called, but fails on the second call (the method overwrites itself with an array). 

When running chains, you will most likely only call `sigma8_0()` once per perturbation instance, which is why you might never run into this issue. Nevertheless, renaming the `self.sigma8_0` array inside the `sigma8_0() `method to just `self.sigma8` would be safer to avoid any issues (and also more accurate, since that array contains a range of sigma8 values, not just the one at z=0).

Please let me know if you need any further checks or clarifications from my side.

gcanasherrera:
@raphkou could you review and investigate this PR?

raphkou:
Hello @NastassiaG, thanks a lot for the PR! I was able to reproduce the bug and test your fix. Could you just update the branch with latest main before I merge it?

Many thanks!

NastassiaG:
Hi @raphkou this should be ready to merge now, please let me know if you need anything else from my side!

NastassiaG:
Hi @raphkou just following up on the discussion on #575 . I can modify the function to remove the reassignment of `self.params_hm_emu["z"]`. Do I understand correctly that this also makes the `max_len` part (including the for loop) unnecessary?

raphkou:
Hi @NastassiaG, yes indeed the whole loop becomes unnecessary since we do not change the size of the redshift array anymore. Thanks for your help!

NastassiaG:
@raphkou I've removed the respective lines, let me know if anything else is needed!

raphkou (approved):

raphkou:
@all-contributors please add @NastassiaG for bug

PR #569, opened 2026-09-09
gh-373: refactoring for tatt and nonlinear bias

### LINKED ISSUE #373: Refactoring for TATT and NonLinear Bias
## Context
Currently, the code computes angular power spectra $C_\ell^{XY}$ using a "summed-kernel" optimization within the Limber approximation.

A single `Tracer` object typically encapsulates multiple physical contributions. For example:
* **Shear Tracer:** Contains a pure lensing contribution ($W_\gamma$) and an Intrinsic Alignment contribution ($W_{IA}$).
* **Number Counts Tracer:** Contains a galaxy density contribution ($W_n$) and a magnification bias contribution ($W_\mu$).

Currently, we assume all these terms scale linearly with the *same* matter power spectrum $P_{\rm m}(k, z)$. This allows the `Tracer` object to sum these contributions internally ($W_{\rm total} = W_{\rm lens} + W_{\rm IA}$) before passing a single kernel to the integrator.

## The Problem
This "summed-kernel" assumption fails for advanced models like **TATT (Tidal Alignment and Tidal Torquing)** or **non-linear galaxy bias**.

In these models, different components of a single tracer couple to different spectral source terms. They can no longer share a common $P_{\rm m}(k)$.
* **Example (TATT):** The lensing component $W_\gamma$ couples to the density field $P_{\delta\delta}$, but the intrinsic alignment component $W_{IA}$ requires tidal field terms (e.g., $P_{\delta E}, P_{EE}$) or cross-terms.
* **Example (Non-linear Bias):** The linear bias term scales with $P_{lin}$, but higher order terms (like $b_2$) couple to loop integrals or separate operators.

**Consequence:**
We must abandon the pre-summation optimization. We need to expand the product of two tracers into their constituent pairs (e.g., Lens-Lens, Lens-IA, IA-IA) and integrate each pair against its specific required power spectrum.

## Proposal: `PowerSpectrumSetup` and Component-wise Integration

We need a refactor that separates the definition of tracer components from the computation of their required spectra.

### 1. Refactor Tracers
`Tracer` objects will no longer return a single summed array. They must expose a list of their physical components.
* `Tracer.get_components()` $\rightarrow [ (W_\text{lens}, \text{Type}_\text{Lens}), (W_\text{IA}, \text{Type}_\text{IA}) ]$

### 2. Introduce `PowerSpectrumSetup` Class
To avoid recomputing power spectra for every integral or guessing which spectra are needed, we introduce a setup phase.
* **Input:** A list of all `Tracer` objects involved in the analysis.
* **Logic:** The `PowerSpectrumSetup` iterates through the tracer components to determine the **union** of all required spectral quantities.
    * *Example 1 (Linear Bias + NLA):* Detects only Density and Linear IA terms $\rightarrow$ Computes only $P_{\rm m}(k)$.
    * *Example 2 (TATT):* Detects Tidal terms $\rightarrow$ Computes $P_{\delta\delta}, P_{\delta E}, P_{EE}$.
* **Output:** A registry/cache containing only the specific $P(k,z)$ arrays needed for this specific run, computed once.

### 3. The Compute Loop
The `compute_cl` function performs the expansion sum. For Tracers $X$ and $Y$, the total $C_\ell$ becomes a sum over all pairs of valid components:

$$
C_\ell^{XY} = \sum_{i \in X} \sum_{j \in Y} \int \frac{d\chi}{\chi^2} \, W_X^i(\chi) \, W_Y^j(\chi) \, P(\text{type}_i, \text{type}_j) (k_\ell,x)
$$

### Benefits
* **Correctness:** Enables TATT and non-linear bias models by linking specific window functions to specific source terms.
* **Efficiency:** The `PowerSpectrumSetup` ensures we only compute the expensive 3D power spectra components strictly required by the current set of tracers.
* **Modularity:** The integrator logic remains generic; the physics of "which component needs which spectrum" is encapsulated in the Setup class.

### Mathematical Justification: The Factorization Assumption

This refactor relies on a structural property of perturbative models (TATT, EFTofLSS, Non-linear Bias): the total power spectrum can be decomposed into a weighted sum of fixed "basis spectra."

We assume the 3D power spectrum between any two tracer components $X$ and $Y$ (e.g., Galaxy Density and Intrinsic Alignment) can be factorized as:

$$
P_{XY}(k, z) = \sum_{n} f_n(\vec{\theta}, z) \cdot P_{\text{basis}, n}(k, z)
$$

Where:

* **$P_{\text{basis}, n}(k, z)$ (The Components):** These are the **Cosmology-Dependent Basis Spectra** (e.g., $P_{\delta \delta}$, $P_{\delta s^2}$, $P_{\delta^2 \delta^2}$). They are computationally expensive to generate (requiring Boltzmann codes or FastPT integrals) but depend *only* on the fundamental cosmological parameters.
* **$f_n(\vec{\theta}, z)$ (The Weights):** These are **Nuisance-Dependent Coefficients**. They are algebraic combinations of the tracer's nuisance parameters (e.g., bias $b_1, b_2$, IA amplitudes $A_1, A_2$) and background growth functions (e.g., $D(z)$). These are computationally cheap to evaluate.

**Implication for Architecture:**
This factorization justifies the proposed `PowerSpectrumSetup` class.
1.  We identify and compute the required unique $P_{\text{basis}, n}$ terms for a given cosmology.
2.  When ecomputing angular spectra, we do **not** recompute the 3D spectra. We simply compute the scalar weights $f_n$ and use them to weight the spectra in integration loops.

This effectively decouples the expensive 3D physics (FastPT/FFTLog) from the nuisance parametes.

(tagging @gcanasherrera @DavidNavarroG @davidesciotti @annmapo @CarmelitaCarbone @martincrocce @itutusaus , to discuss whether the abstraction I am makin might work!)

gcanasherrera:
### ✅ Summary

Adds TATT (Tidal Alignment + Tidal Torquing) intrinsic-alignment modeling to `ShearTracer`, via a new generalized, opt-in `Contribution`/spectrum-engine architecture based on the toy from @marcobonici that leaves every existing (non-TATT) code path and public signature unchanged. Includes both an illustrative kernel backend (no extra dependency) and a real one loop backend (`fast-pt`, matching how TATT was implemented in production CLOE). Also reorganizes `cloelib.observables` into `photo/`/`spectro/` packages so related tracer/backend code lives together, and brings docs/tests up to date with the final (non-phased) architecture.

### 🔄 Changes

- **Contribution-based tracer decomposition**: `ShearTracer`/`PositionsTracer` window functions are now expressed as a sum over `Contribution` objects (`get_contributions()`), bit-for-bit identical to the previous fused implementation for every existing configuration (`tests/test_photo_tracers_characterization.py`, `tests/test_contributions.py`).
- **Generalized per-contribution-pair spectrum engine** (`cloelib/observables/photo/spectrum_engine.py`): lets a `Contribution` declare extra derived power spectra it needs beyond the plain matter Pk, computed once and only if needed. `AngularTwoPoint.get_Cl` forks on a cheap static check (`needs_generalized_engine`) — `False`, and the original fast path runs unmodified, for every tracer built with today's defaults.
- **TATT intrinsic-alignment model**: new `ia_model="TATT"` option on `ShearTracer`, implementing Eqs. (9)-(16) of Navarro-Gironés et al. (2026, arXiv:2602.16448) / Blazek et al. (2019). Declares 10 one-loop kernel requests for the II term, pruned to 4 for the GI term.
- **Two TATT kernel backends**, selected via a new `tatt_loop_computer` kwarg:
  - `PlaceholderTATTLoopComputer` (default) — illustrative, no extra dependency.
  - `PBJTATTLoopComputer` — real physics via `fastpt.FASTPT.IA_ta`/`.IA_tt`/`.IA_mix` (new optional extra: `pip install cloelib[fastpt]`), matching production CLOE's implementation term-for-term. Resolves the required *linear* power spectrum directly from the tracer's own `perturbations` object (via `.linearperturbations`, now also set by `HMemuNonLinearPerturbations`).
- **Two correctness/robustness fixes surfaced along the way**:
  - `ShearTracer.get_window_IA`'s `growth_factor(...)[:, 1]` handling made robust across backends (JAX backends return a 1D D(z) with no `.k` attribute).
  - `Pkl_interp_signed` (signed-spectrum Limber interpolation) now clamps queries to the k-grid's own domain instead of extrapolating — a steep, sign-changing one-loop kernel could otherwise overflow to `inf`/`NaN` at the low-χ edge of the grid.
- **Module reorganization**: `cloelib/observables/photo.py` → `photo/` package (`shear.py`, `positions.py`, `tracer.py`, `contributions.py`, `spectrum_engine.py`); `spectro.py`/`PBJ_spectro.py`/`CometEFT_spectro.py`/`CometVDG_spectro.py` → `spectro/` package. Public imports (`from cloelib.observables.photo import ShearTracer, PositionsTracer`, `from cloelib.observables.spectro import SpectroPower`) are unchanged; every other import site (internal modules, tests, tutorial notebooks) updated to the new paths — no back-compat shims left behind.
- **Docs updated**: `CONTRIBUTION_ARCHITECTURE.md` (replaces the stale `CONTRIBUTION_ARCHITECTURE_PROPOSAL.md`) describes the final architecture; `docs/api.md` and `docs/code_structure/observables/*.md` reflect the new module layout and document `ia_model="TATT"`; `paper/paper.md` example import fixed.
- **Tests renamed and cleaned**: `test_phase0_characterization.py` → `test_photo_tracers_characterization.py`, `test_contributions_phase1.py` → `test_contributions.py`, `test_tatt_phase2.py` → `test_tatt.py`, with all "Phase 0/1/2" development-journal language removed from docstrings/comments throughout the source tree.

### 🛠 How to Test

```bash
pip install -e .                    # base install
pip install -e .[fastpt]       # optional: real TATT kernels via fast-pt

pytest tests/
Targeted:


pytest tests/test_photo_tracers_characterization.py  # golden get_window/get_Cl regression
pytest tests/test_contributions.py                    # Contribution decomposition == old fused methods
pytest tests/test_tatt.py                              # TATT architecture + (if fast-pt installed) real backend
```

A worked, executed example (ia_model="TATT" with both backends, plotted against NLA) will be made available in the sibling playground repo photo notebook: tutorials/observables/photo.ipynb.

### 📝 Documentation
 This PR updates documentation
 
### 🏗 Related Issues
Resolves #373 

7b9e3229-627d-4ab6-9e7f-fde6d3601423

📌 Additional Notes
Known, documented limitations (not fixed in this PR, flagged in CONTRIBUTION_ARCHITECTURE.md §3):

- CAMBNonLinearPerturbations doesn't yet expose .linearperturbations; passing one to PBJTATTLoopComputer would silently use the nonlinear Pk as FAST-PT input. (volunteer?)
- TATT's real B-mode power (P_II^BB) is computed internally but not yet wired into AngularTwoPoint's SHE-SHE output packaging, which still hardcodes B-mode slots to zero. (what do we want to do with this?)
- PositionsTracer(include_rsd=True) paired with a generalized-engine contribution raises NotImplementedError rather than silently producing wrong physics

---

- [x] I have updated documentation (if applicable) - docstrings & instructions in `cloelib/docs/`

- [ ] Check that the documentation has been updated accordantly (docstrings & instructions in `cloelib/docs/`)


marcobonici:
Hi @gcanasherrera ,
as I already wrote, I am mostly happy about this PR!

I just aìhave a few comments.

1. Pkl_interp_signed. I clearly see, mathematically, why we have to introduce it (as we now have spectra that do a zero-crossin) but it is worth adding a TODO to include some additional tests in the future (just scared about gradients behaviour).
2. Multiplicative shear. The generalized path bypasses `ShearTracer.get_window,` which is currently where multiplicative shear calibration is applied. Could we apply the tracer-level calibration explicitly to the generalized result, and add nonzero-\(m_i\) regression tests for SHE–SHE and POS–SHE?
3. Potentially missing Linear pertubrations. I think this is worth making safer in the current PR rather than leaving it only as a TODO.

At the moment, `PBJTATTLoopComputer` effectively does:

```
self.linear_perturbations = getattr(
    perturbations, "linearperturbations", perturbations
)
```

The problem is that the absence of a .linearperturbations attribute does not imply that the supplied object is itself linear.

This can therefore fail silently for nonlinear backends that do not expose their corresponding linear perturbations object.

For example, consider:

```
perturbations = CAMBNonLinearPerturbations(...)
loop_computer = PBJTATTLoopComputer(perturbations)
```

If CAMBNonLinearPerturbations does not expose .linearperturbations, the current fallback sets

self.linear_perturbations = perturbations

and FAST-PT is then fed the nonlinear CAMB matter power spectrum as if it were linear. The code still runs and may return perfectly finite spectra, which makes this particularly dangerous: the failure is physical rather than numerical (and silent errors can be more problematic...).

I would prefer that we fail explicitly whenever a trusted linear source cannot be identified.

Conceptually:

```
linear = getattr(perturbations, "linearperturbations", None)

if linear is not None:
    self.linear_perturbations = linear
elif is_known_linear_perturbations(perturbations):
    self.linear_perturbations = perturbations
else:
    raise ValueError(
        "PBJTATTLoopComputer requires a linear matter-power-spectrum source. "
        "Pass a linear perturbations object directly, or use a nonlinear "
        "backend exposing its corresponding `.linearperturbations` object."
    )
```

The exact way we determine whether an object is a linear Perturbations backend can follow the conventions already used in cloelib; the important point is that we should not infer linearity merely from a missing attribute.

I donot think this PR need to add full support for every nonlinear backend. In particular, fixing the CAMB interface can remain future work. I would just make unsupported configurations fail immediately now.

A follow-up TODO could then be:

```
# TODO: standardize access to the corresponding linear perturbations
# object across nonlinear cosmology backends, so PT-based observables
# can retrieve a linear P(k,z) through a common interface.
```

That would give us safe behavior now, while leaving the cleaner backend-level abstraction for a dedicated future task.

gcanasherrera:
@marcobonici back to you after your comments above. I will open backlog issues tomorrow morning for the TODOs.

marcobonici (approved):
We are now good to go IMHO :)

